Issue #13, opened 2016-09-22
how to disable log on release app?


li9ht:

passsy:
There is definitely room for improvement. For now you can override `TiActivity#logTiMessages` and add a custom behavior

```java
    @Override
    public void logTiMessages(final String msg) {
        if (BuildConfig.DEBUG) {
            // only log for debug builds
            super.logTiMessages(tag);
        }
    }
```

StefMa:
@li9ht we have fixed this issue. 
See #14 

Now you are able to use a custom logger (or no logging by default):
```java
Timber.plant(new Timber.DebugTree());
TiLog.setLogger(new TiLog.Logger() {
    @Override
    public void log(final int level, final String tag, final String msg) {
        Timber.tag(tag).log(level, msg);
    }
});
```

PR #14, opened 2016-09-26
Use the same logging interface everywhere

passsy:
`TiLog.setLogger(null)`
- allows custom loggers
- disable logging is now possible

Fixes logging with `java.util.logging.Logger` not being visible because the where printed with log level `FINE` which is not printed when the library isn't in debug mode.

issues #13 

combine with Timber:

```java
Timber.plant(new Timber.DebugTree());
TiLog.setLogger(new TiLog.Logger() {
    @Override
    public void log(final int level, final String tag, final String msg) {
        Timber.tag(tag).log(level, msg);
    }
});
```

### DIFF
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiLog.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiLog.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright (C) 2016 grandcentrix GmbH
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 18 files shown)


passsy:
Not sure if the logger should be `null` by default.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiLog.java
```diff
@@ -0,0 +1,106 @@
...
+         *              {@link Log#WARN},{@link Log#ERROR}
+         * @param tag   log tag, caller
+         * @param msg   message to log
+         */
+        void log(final int level, final String tag, final String msg);
+    }
+
+    private static final String TAG = "ThirtyInch";
+
+    // logs everything to Timber by default
```
StefMa: This is wrong since https://github.com/grandcentrix/ThirtyInch/pull/14/commits/4da934b2989a0fbc99120cc421677641cc87a21b

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiLog.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public static void w(final String tag, final String msg) {
+        if (logger != null) {
+            logger.log(Log.WARN, tag, msg);
+        }
+    }
+
+    TiLog() {
+        throw new AssertionError("no instances");
```
StefMa: Why not just setting constructor to private?

StefMa:
LGTM 👍

PR #15, opened 2016-09-28
cleanup logging

passsy:
- removed lifecycle logging (which was only important during development)
- adjusted log level. Most is no verbose
- improved `TiFragment#onDestroy` loggig analog to `TiActivity#onDestroy`


StefMa:
LGTM

PR #19, opened 2016-10-05
Use equals() instead of hashCode() for comparing arguments

fzymek:
Use equals() for comparing arguments instead of using hashCode() in `DistinctUntilChangedInvocationHandler`.

Update grade plugin to latest version and do small code cleanup

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.3'
+        classpath 'com.android.tools.build:gradle:2.2.0'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.0'
+        classpath 'com.android.tools.build:gradle:2.1.3'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
@@ -33,11 +33,11 @@ final class DistinctUntilChangedInvocationHandler<V> extends AbstractInvocationH
 
     private final V mView;
 
-    DistinctUntilChangedInvocationHandler(V view) {
+    public DistinctUntilChangedInvocationHandler(V view) {
         mView = view;
     }
 
-    void clearCache() {
+    public void clearCache() {
         mLatestMethodCalls.clear();
     }
 
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
@@ -28,15 +28,15 @@ final class DistinctUntilChangedInvocationHandler<V> extends AbstractInvocationH
 
     private static final String TAG = DistinctUntilChangedInvocationHandler.class.getSimpleName();
 
-    private HashMap<String, Integer> mLatestMethodCalls = new HashMap<>();
+    private HashMap<String, Object[]> mLatestMethodCalls = new HashMap<>();
 
     private final V mView;
 
-    public DistinctUntilChangedInvocationHandler(V view) {
+    DistinctUntilChangedInvocationHandler(V view) {
         mView = view;
     }
 
-    public void clearCache() {
+    void clearCache() {
         mLatestMethodCalls.clear();
     }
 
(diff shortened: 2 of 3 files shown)


passsy:
Thanks for your work. The reason for using `hashcode()` instead of `equals()` is that the InvocationHandler should not hold references to the actual objects sent to the view. I'd be ok when you use a weak reference.

Please keep PRs clean and change build stuff when not required.

fzymek:
Hmm, right. I haven't thought of it. I'll follow you idea and see how it works with WeakRefs

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.1.3'
```
passsy: Please don't change the gradle version in the PR

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
```diff
@@ -28,15 +29,15 @@
...
 
     private final V mView;
 
-    public DistinctUntilChangedInvocationHandler(V view) {
+    DistinctUntilChangedInvocationHandler(V view) {
         mView = view;
     }
 
-    public void clearCache() {
+    void clearCache() {
```
passsy: Don't change the visibility. `clearCache()` is part of the public API and should remain there

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
```diff
@@ -28,15 +29,15 @@
 
     private static final String TAG = DistinctUntilChangedInvocationHandler.class.getSimpleName();
 
-    private HashMap<String, Integer> mLatestMethodCalls = new HashMap<>();
+    private HashMap<String, WeakReference<Object[]>> mLatestMethodCalls = new HashMap<>();
 
     private final V mView;
 
-    public DistinctUntilChangedInvocationHandler(V view) {
+    DistinctUntilChangedInvocationHandler(V view) {
```
passsy: Don't change the visibility. `DistinctUntilChangedInvocationHandler` is part of the public API and should remain there

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
```diff
@@ -28,15 +29,15 @@
 
     private static final String TAG = DistinctUntilChangedInvocationHandler.class.getSimpleName();
 
-    private HashMap<String, Integer> mLatestMethodCalls = new HashMap<>();
+    private HashMap<String, WeakReference<Object[]>> mLatestMethodCalls = new HashMap<>();
```
passsy: Use a `WeakHashMap<String, Object[]>` instead of `HashMap<String, WeakReference<Object[]>>`

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
fzymek: Ahh, aggresive auto formating is evil

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
fzymek: I do not see how `WeakHashMap` is benefit here, since it allows keys to be garbage collected. Since we put method names as keys and method arguments as values and I doubt our method name will be gc'ed.

fzymek:
After giving a second thought, this PR might be unnecessary. We do not gain much by using `equals` and using WeakRefs makes code harder to understand/maintain. As long as developers will honour `hashCode()` and `equals()` contract, current solution should be safe.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
passsy: my bad, I was assuming the value is referenced weak

passsy:
I still think this PR is valuable since in theory two different arguments could have the same `hashcode`

fzymek:
hmm, ok, I'll keep working on it and maybe I'll find better solution

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandler.java
fzymek: restating my deleted comment:

Since `WeakHashMap` treats keys as objects that can be garbage collected, and here we use method names as keys and method arguments as values, I don't think using `WeakHashMap ` will be a gain here

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedInvocationHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+        Object[] args = new Object[]{"p1"};
+
+
+        //when
+        handler.handleInvocation(null, method, args);
+        handler.handleInvocation(null, method, args);
+        //noinspection UnusedAssignment
+        args = null;
+        //try to clear references
+        System.gc();
```
fzymek: `clearCache()` deletes all entries from map which makes this test unneccesary in this case

fzymek:
Ok, for now I've restored all cleared public API and gradle plugin version. I've also added `clearCache()` call in test as suggested.

passsy:
Thanks so far. I have to think about the consequences. Here my thoughts:

- `hashcode` obviously was not correct because in theory two different arguments could have the same hashcode
- holding strong references inside a invocation handler seems wrong. What if one sends big blobs (images) to the view? Having control is important
- In the past I've sent immutable objects from my presenter to the view created in Observable streams. I don't hold references to those. Those kind of objects will be garbage collected quite often. This wasn't a problem with the hashcode implementation since the hashcode was the same. Now the view would be called twice with the same data because the `InvocationHandler` can't compare the new arguments to the old garbage collected arguments.

On idea is to add an parameter to the annotation like `@DistinctUntilChanged(weakComparison = true)`

Any thought on this @jannisveerkamp @StefMa?

jannisveerkamp:
I guess it's impossible to get the desired behavior without saving the whole object (strong reference!).

I like the idea of choosing how the comparison is done. Have to think about it further.

PR #20, opened 2016-10-05
Add TiDialogFragment

remcomokveld:
Speaks for itself. Although I can think of prettier, delegate based, solutions to add MVP support for DialogFragments this is a simple copy from TiFragment but it extends android.support.v4.app.DialogFragment instead of android.support.v4.app.Fragment

### DIFF
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiDialogFragment.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiDialogFragment.java
@@ -1,0 +1,260 @@
+package net.grandcentrix.thirtyinch;
+
+import android.os.Bundle;
+import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
+import android.support.v4.app.DialogFragment;
+import android.support.v4.app.FragmentActivity;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+
+import net.grandcentrix.thirtyinch.callonmainthread.CallOnMainThreadInterceptor;
+import net.grandcentrix.thirtyinch.distinctuntilchanged.DistinctUntilChangedInterceptor;
+import net.grandcentrix.thirtyinch.internal.InterceptableViewBinder;
+import net.grandcentrix.thirtyinch.internal.PresenterSavior;
+import net.grandcentrix.thirtyinch.internal.PresenterViewBinder;
+import net.grandcentrix.thirtyinch.internal.TiLoggingTagProvider;
+import net.grandcentrix.thirtyinch.internal.TiPresenterProvider;
+import net.grandcentrix.thirtyinch.internal.TiViewProvider;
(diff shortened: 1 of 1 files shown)


passsy:
please have a look at this @lacky1991

StefMa:
Don't know if i like the idea.
Have you really so much business logic inside your Dialog? 🤔 
However. If we do it, then we need to use a Delegate and merge all the duplicated code from TiFragment and TiDialogFragment...

passsy:
Yes, a delegate is definitely required. This is also required when we want to support `TiFragment` and `TiDialogFragment` for the `plugin`.
As a first step I'm ok with the duplicated code. But I remember @lacky1991 found a edge case when he implemented `TiDialogFragment` himself.

remcomokveld:
If I have a dialog where I can edit a data object and what to save/sync this changes,  I need to have a tested presenter which does that business logic. Totally agree with the fact that it should be implemented with a delegate

StefMa:
@remcomokveld are you still working on it?
It seems that more people needs the DialogFragment than I thought 😃

Nxt3:
I noticed that while using this the layout of my Dialog gets wonky. If I extend `DialogFragment` instead of `TiDialogFragment`--the layout is fine. Not sure what the cause could be.

passsy:
Looks fine for me with the `TiDialogFragment`.
screenshot-2016-11-05_15 37 00 507

And I can't imagine the code of `TiDialogFragment` causes this. My best guess is some caching bug of instant run 😉

Nxt3:
I don't use instant run. It seems to be the difference of extending "DialogFragment" vs "AppCompatDialogFragment"

passsy:
Now I got it! Yes the `TiDialogFragment` should extend `AppCompatDialogFragment` and not `DialogFragment`.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiDialogFragment.java
```diff
@@ -0,0 +1,260 @@
...
+import net.grandcentrix.thirtyinch.internal.TiLoggingTagProvider;
+import net.grandcentrix.thirtyinch.internal.TiPresenterProvider;
+import net.grandcentrix.thirtyinch.internal.TiViewProvider;
+import net.grandcentrix.thirtyinch.util.AndroidDeveloperOptions;
+import net.grandcentrix.thirtyinch.util.AnnotationUtil;
+
+import java.util.List;
+
+public abstract class TiDialogFragment<P extends TiPresenter<V>, V extends TiView>
+        extends DialogFragment implements TiPresenterProvider<P>, TiLoggingTagProvider,
```
passsy: should extend `AppCompatDialogFragment` instead of `DialogFragment`

StefMa:
It seems @remcomokveld don't maintain this anymore.
I've just created another PR with the AppCompat fix. See #42.

This will be closed.

PR #23, opened 2016-10-10
Use provided instead of compile for support libraries

passsy:
ThirtyInch should not increase the version of the support library when added to a project. This change prevents necessary exclusions of AppCompat when the app project uses an older version than ThirtyInch

```java
// exclusion not required anymore    
compile("net.grandcentrix.thirtyinch:thirtyinch:$thirtyinchVersion") {
    exclude group: "com.android.support", module: "appcompat-v7"
}
```

Since every project uses AppCompat this should make problems. If a project doesn't have the dependency yet it must be added manually (`compile "com.android.support:appcompat-v7:$supportLibraryVersion"`)

Same behavior for CompositeAndroid. The plugin artifact won't ship the compositeAndroid dependency. Instead it will use the CompositeAndroid version of the app project. This change prevents random internal errors when CompositeAndroid and AppCompat use different versions


StefMa:
I'm with you.
But this don't solve this problem
> ThirtyInch should not increase the version of the support library when added to a project.

I'm with you that a lib shouldn't increate another dependency from a project.
But a project should change the dependency from a dependency 😕  
It will be very hard work to figure out bugs when every project which uses `ThirtyInch` use a different `SupportLib` which have directly influence to `ThirtyInch`...

passsy:
If you print the dependencies of some random project `./gradlew :app:dependencies` you'll see a lot of dependencies of libraries getting automatically increased. This is normal dependency resolution.

For example here:
```
+--- uk.co.chrisjenx:calligraphy:2.2.0
|    \--- com.android.support:appcompat-v7:23.1.1 -> 24.1.1 (*)
```
This dependency upgrade is ok, and works most of the time. If it doesn't you can solve it with an `exclude` rule.

In the past the code of `ThirtyInch` never had to be adjusted when upgrading the `SupportLib` starting with `v19`. I'm confident it wont break for future updates.

StefMa:
Talked about that internally.
Everything is fine here 👍

PR #24, opened 2016-10-11
DistinctUntilChanged with different comparisons

passsy:
As discussed in #19 the current implementation of `@DistinctUntilChanged` using the `hashCode()` of the arguments could cause problems.

The new default implementation uses `equals()` instead of `hashcode()`. A `DistinctComparator` can be set as argument of the annotation `@DistinctUntilChanged(comparator = EqualsComparator.class)`. Three different implementations exist:
- `EqualsComparator` (default) - uses `equals()` for comparison and holds a strong reference to the previous arguements
- `WeakEqualsComparator` -  uses `equals()` for comparison and holds a weak reference to the previous arguments. Could call a method twice when `gc()` kicks in.
- `HashComparator` - old implementation using `hashCode()` for comparison. Doesn't hold references to the arguments

When writing tests I detected that the `DistinctUntilChangedInvocationHandler` swallows exceptions thrown by the `TiView`. The same problem was already discovered when using the `@CallOnMainThread` annotation 23034a0b98c2ed2a8548b0ed1f7575371188de92. I fixed it here accordingly

These proguard rules are required:

```
# ThirtyInch
-keep class net.grandcentrix.thirtyinch.distinctuntilchanged.**
-keep class net.grandcentrix.thirtyinch.callonmainthread.**
```

### DIFF
--- a/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedTest.java
+++ b/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedTest.java
@@ -1,0 +1,194 @@
+package net.grandcentrix.thirtyinch.distinctuntilchanged;
+
+import net.grandcentrix.thirtyinch.TiView;
+
+import org.junit.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsEqual.equalTo;
+
+/**
+ * Testing {@link DistinctUntilChanged} annotation via {@link DistinctUntilChangedInterceptor}
+ * requires interfaces extending {@link TiView}.
+ */
+public class DistinctUntilChangedTest {
+
+    private class CountWrapper {
+
+        private int mCalled = 0;
+
+        public void call() {
+            mCalled++;
+        }
+
+        public int getCalled() {
+            return mCalled;
+        }
+    }
+
+    /**
+     * This class doesn't have a good hashcode method
+     */
+    private class BadHashObject {
+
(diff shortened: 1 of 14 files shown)


thirtyinch/src/test/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedTest.java
```diff
@@ -0,0 +1,194 @@
...
+
+    private interface TestViewEquals extends TiView {
+
+        @DistinctUntilChanged(comparator = EqualsComparator.class)
+        void annotatedMethod(String str);
+    }
+
+    private interface TestViewBadHash extends TiView {
+
+        @DistinctUntilChanged(comparator = EqualsComparator.class)
```
StefMa: Name and comparator isn't equal. You say `BadHash` (and argument is `BadHash`) but using `EqualsComparator`

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/distinctuntilchanged/DistinctUntilChangedTest.java
```diff
@@ -0,0 +1,194 @@
...
+        void annotatedMethod(BadHashObject bho);
+    }
+
+    private interface TestViewBadEquals extends TiView {
+
+        @DistinctUntilChanged(comparator = HashComparator.class)
```
StefMa: Name and comparator isn't equal. You say `BadEquals` (and argument is `BadEquals`) but using `HashComparator `

StefMa:
Rest LGTM

passsy:
@StefMa Thanks for pointing out the mistake in the tests!

PR #26, opened 2016-10-14
Change onWakeUp() to onAttachView(TiView) and onSleep() to onDetachView()

passsy:
The new naming is more convenient throughout most android MPV libraries. Switching to `Ti` should now be simpler. It reflects better that it's only about the view being absent. wakeup and sleep could imply the Presenter is kind of starting/stopping which is not the case.

`onAttachView(TiView)` has the attached view as first parameter which is `@NonNull`. No calls to `getView()` are neccessary inside `onAttachView(TiView)`.
`getView()` is now `@Nullable` (and always was null when not between `wakeup()` and `sleep()` state) and causes lint to hint a Nullable warning. This warning was missing and caused a lot of NPE for `Ti` beginners

### DIFF
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
@@ -20,6 +20,7 @@ import net.grandcentrix.thirtyinch.internal.OneTimeRemovable;
 
 import android.app.Activity;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
 import android.support.v4.app.Fragment;
 
@@ -44,16 +45,16 @@ public abstract class TiPresenter<V extends TiView> {
         INITIALIZED,
         /**
          * presenter is running fine but has no attached view. Either it gets a view  and
-         * transitions to {@link #VIEW_ATTACHED_AND_AWAKE} or the presenter gets destroyed ->
+         * transitions to {@link #VIEW_ATTACHED} or the presenter gets destroyed ->
          * {@link
          * #DESTROYED}
          */
-        CREATED_WITH_DETACHED_VIEW,
+        VIEW_DETACHED,
         /**
          * the view is attached. In any case, the next step will be {@link
-         * #CREATED_WITH_DETACHED_VIEW}
+         * #VIEW_DETACHED}
          */
-        VIEW_ATTACHED_AND_AWAKE,
+        VIEW_ATTACHED,
         /**
          * termination state. It will never change again.
          */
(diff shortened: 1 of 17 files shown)


thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
```diff
@@ -155,12 +151,37 @@ public void bindNewView(@NonNull final V view) {
...
+            }
+        }
+
+        if (!isInitialized()) {
+            throw new IllegalStateException("Presenter is not created, call #create() first");
         }
 
         mView = view;
+        moveToState(State.VIEW_ATTACHED, false);
+        mCalled = false;
```
StefMa: mCalled isn't a good name.
I think it would be nicer if we have two booleans. `mViewAttachedCalled` and `mViewDetachedCalled`.
So we can't override the single boolean anymore...

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
```diff
@@ -288,25 +323,18 @@ public String toString() {
     }
 
     /**
-     * when calling {@link #wakeUp()} the presenter can start communicating with the {@link
-     * TiView}.
+     * The view is now attached and ready to receive events. {@link #getView()} is not guaranteed
```
StefMa: Better Javadoc:
`getView() may be null`.
Current javadoc is very confused and have to be read more than once ;)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
```diff
@@ -342,26 +370,34 @@ protected void onDestroy() {
      * Right after this method the view will be detached. {@link #getView()} will return
      * <code>null</code> afterwards.
      *
-     * @see #sleep()
+     * @see #detachView()
      */
+    protected void onDetachView() {
+        if (mCalled) {
+            throw new IllegalAccessError("don't call #onSleep() directly, call #detachView()");
```
StefMa: > don`t call #onDetachView () directly... 

right?

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiPresenterTest.java
```diff
@@ -37,6 +39,104 @@
...
+        } catch (IllegalStateException e) {
+            assertTrue(e.getMessage().contains("terminal state"));
+        }
+    }
+
+    @Test
+    public void attachWithoutInitialize() throws Exception {
+
+        // not calling
+        // mPresenter.create();
```
StefMa: Can be removed. The method should say what happen here.
Maybe a better name is `attachWithoutOnCreate`. But I'm also fine with the current name

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
passsy: it's a general property for all super call checks. Implementation and naming copied from AOSP `Actiivty` and `Fragment` https://github.com/android/platform_frameworks_base/blob/4b1a8f46d6ec55796bf77fd8921a5a242a219278/core/java/android/app/Activity.java#L748

PR #36, opened 2016-11-07
Queue actions send to the view until it is attached

passsy:
Sending events to the view although the view isn't attached is already possible with rx and `.compose(rxUtils.deliverLatestCacheToView(this))` which caches all items until the view is attached and emits them in the same order. The same behavior is now back ported to pure java.

*Usecase*: Building a splashscreen which should make a network call and then forward to a new `Activity`.
- It should not start the activity when the app was closed in the meantime
- It should start the new activity when the app comes to the foreground and the network call already finished

The network call is started in `TiPresenter#onCreate()`. The call can return at any time. 
- Before the view is attached when a cached result is returned. 
- When the view is attached. 
- When the user left the screen and the view is detached.

The call `view.goToMain()` will only be sent once to the view when the activity is open (view is attached) after the network call finished. There will be no `NPE` because the view is `null`.

```java
// SplashscreenPresenter
@Override
protected void onCreate() {
    super.onCreate();
    doNetworkCall(result -> {
        handleResult(result);
        sendToView(view -> view.goToMain());
    });
}

// SplashscreenActivity
@Override
public void goToMain() {
    startActivity(new Intent(this, MainActivity.class));
    finish();
}
```


oleksandrbogdanov:
The storing of results logic is clear, and tests is passed I would approve it. Do not have permissions to merge.

Issue #37, opened 2016-11-10
Need method  manageSubscriptions(Subscription ...subscription) in RxTiPresenterSubscriptionHandler


ar-g:
For conciseness, it would be helpful to have method  manageSubscriptions(Subscription ...subscription) in RxTiPresenterSubscriptionHandler

passsy:
Can you give me an example where this is useful?

ar-g:
The point is to not write manageSubscription() 10 times in presenter in onAttachView. Though could be easily achieved manually but would be just nice to have such method in RxTiPresenterSubscriptionHandler.

StefMa:
Fixed in #43 
Will be shipped with the next release

PR #43, opened 2016-12-13
Add manageViewSubscriptions and manageSubscriptions 

StefMa:
Add manageViewSubscriptions and manageSubscriptions to handle multiple subscriptions at once.

Feature request of #37

### DIFF
--- a/rx/src/test/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandlerTest.java
+++ b/rx/src/test/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandlerTest.java
@@ -25,8 +25,10 @@ import org.junit.runners.JUnit4;
 
 import rx.observers.TestSubscriber;
 
+import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.core.IsEqual.equalTo;
+import static org.junit.Assert.fail;
 import static org.mockito.Mockito.mock;
 
 @RunWith(JUnit4.class)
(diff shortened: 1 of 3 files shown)


rx/src/test/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandlerTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+    @After
+    public void tearDown() throws Exception {
+        mPresenter = null;
+        mView = null;
+        mSubscriptionHandler = null;
+    }
+
+    @Test(expected = AssertionError.class)
+    public void tesManageViewSubscription_DetachBeforeAttach_ShouldThrowAssertError()
```
passsy: Expected: `tes[t]ManageViewSub...` but instead
Found: `tesManageViewSub...`

rx/src/test/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandlerTest.java
```diff
@@ -0,0 +1,159 @@
...
+        mPresenter.create();
+        TestSubscriber<Integer> testSubscriber = new TestSubscriber<>();
+
+        mSubscriptionHandler.manageViewSubscription(testSubscriber);
+        mPresenter.detachView();
+
+        testSubscriber.assertUnsubscribed();
+    }
+
+    @Test(expected = IllegalStateException.class)
```
passsy: I prefer the previous used try catch solution because the exception message was checked too

PR #46, opened 2016-12-15
Revert using equals instead of hashcode

passsy:
We found problems in two of our apps when we migrated from `0.7.1` to `0.8.0-rc1`. View methods haven't been called because mutable objects have been used as parameters. After mutating those objects and calling the method again the View method wasn't called as expected. The parameter changed but the `EqualsComparator` is holding a reference to exactly this changed object. The second call simply detects the same object reference and omits forwarding the method call to the view.

After thinking a lot about this I think the hashcode solution is better. It stores the hashcode of the called object at the time it is called. At the next call it compares the current hashcode to the stored hashcode. This also works when the same object was mutated and changes will be detected.

Reverts #19, #24


StefMa:
> This also works when the same object was mutated and changes will be detected.
But I don't get it. Why? 🤔 
The hashCode got changed when the the Objects was changed too, right?

It would be nice if you can make a Test which will reflect exactly this behaviour...

passsy:
@StefMa You understood it correctly. Tests added

PR #47, opened 2016-12-15
Add fragment delegate

vRallev:
Add a fragment delegate and remove duplicated code. This allows reusing the fragment code without extending `TiFragment` or `TiDialogFragment`.


README.md
```diff
@@ -316,6 +316,7 @@ public class HelloWorldActivity extends CompositeActivity implements HelloWorldV
 }
 ```
 
+
```
StefMa: Remove this line

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/TiFragmentDelegate.java
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    @SuppressWarnings("UnusedParameters")
+    public void onCreateView(final LayoutInflater inflater, @Nullable final ViewGroup container,
+            @Nullable final Bundle savedInstanceState) {
+        mViewBinder.invalidateView();
+    }
+
+    public void onDestroy() {
+        //FIXME handle attach/detach state
```
StefMa: Fix this.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/TiFragmentDelegate.java
```diff
@@ -0,0 +1,256 @@
...
+                public void run() {
+                    if (isUiPossible() && mActivityStarted) {
+                        mViewBinder.bindView(mPresenter, mViewProvider);
+                    }
+                }
+            });
+        }
+    }
+
+    public void onStop() {
```
StefMa: In our `TiActivityDelegate` we have called these method with the suffix `_afterSuper`. 
It would be great if you either
* Add this prefix here too.
* Remove the prefix from the `TiActivityDelegate`
It is more consistent...

README.md
StefMa: Remove this line

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/TiFragmentDelegate.java
StefMa: Fix this :)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/DelegatedTiFragment.java
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package net.grandcentrix.thirtyinch.internal;
+
+import android.app.Activity;
+import android.support.v4.app.Fragment;
+
+public interface DelegatedTiFragment {
```
vRallev: I'd say, because of the `P getRetainedPresenter();` method.

rberghegger: Sure we could. Or we could rename all activity related callbacks in the `DelegatedTiFragment` something like: `isHostingActivity...`

passsy: some methods overlap, some don't. I don't see a reason for inheritance

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/TiFragmentDelegate.java
vRallev: I've added the suffix

README.md
vRallev: Done

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/DelegatedTiFragment.java
StefMa: > some methods overlap

Currently only the `P getRetainedPresenter();` don't overlap. The question for me is: Why we haven't it inside the fragment? 

> I don't see a reason for inheritance

passsy: `Fragment` and `Activity` don't share the same interface although many lifecycle methods are equal. Both work differently. Also methods like `isActivityChangingConfigurations()` are currently not used by the Fragment but for logging and may be removed soon.

`P getRetainedPresenter();` is not required for Fragments because the Fragment Object doesn't die and gets retained together with the delegate automatically.

I'm going with the renaming suggested by @rberghegger

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiDialogFragment.java
```diff
@@ -74,144 +65,83 @@ public P getPresenter() {
...
     public void onStop() {
-        mActivityStarted = false;
-        mPresenter.detachView();
+        mDelegate.onStop_beforeSuper();
         super.onStop();
     }
 
+    @Override
+    public boolean postToMessageQueue(final Runnable runnable) {
+        return getActivity().getWindow().getDecorView().post(runnable);
```
passsy: true. It works for us because it's only called in `onStart()`. But sadly this is public API and should always work. I suggest creating a separate PR for this

StefMa:
@vRallev thank you for you contribution!

Issue #39, opened 2016-11-18
Presenters cannot restore instance state after process has died


ihbyun:
It looks like there's no good way to restore the instance state of a Presenter in ThirtyInch when the process is killed and restarted.

The PresenterSavior handles the case when the process is NOT killed by essentially holding a static map of Presenters, but this map will get cleared when the process dies and the Activity gets restored.

This can be reproduced by running the sample code and taking the following steps:
1. Go to Developer Options and turn on Don't Keep activities and set Background Process limit to "No background processes"
2. Install the provided sample apk from https://github.com/grandcentrix/ThirtyInch/tree/v0.7.1/sample
3. Open the app
4. Press Home
5. Navigate back to the app

After following the steps you can see in the logs

11-13 05:49:47.407   909   909 I ThirtyInch: HelloWorldActivity:TiActivity@9cbcc1d: could not recover the Presenter although it's not the first start of the Activity. This is normal when configured as .setRetainPresenterEnabled(false).

As far as I understand setRetainPresenterEnabled is true by default and no one sets it in the sample code.

Is this an oversight or is there a recommended way to take care of this case?

passsy:
ThirtyInch can nothing do against process death and has no APIs to restore the state. It was developed by assuming that the process rarely gets killed which is true for modern phones.
When you see your app process getting killed you have all Android tools to restore the state you already know.

## savedInstanceState

You can save the state of the presenter in `onSaveInstanceState(Bundle)` and restore the state when the presenter gets recreated `providePresenter()` after a process death

```java
public class MyActivity extends TiActivity<MyPresenter, MyView> implements MyView {

    // make it accessible in providePresenter()
    private Bundle mSavedInstanceState = null;

    @Override
    protected void onCreate(final Bundle savedInstanceState) {
        mSavedInstanceState = savedInstanceState;
        super.onCreate(savedInstanceState);
    }

    @Override
    protected void onSaveInstanceState(final Bundle outState) {
        super.onSaveInstanceState(outState);
        // using a byte[] here because the presenter should have zero Android dependencies
        // Bundle is hard to mock
        outState.putByteArray("presenterState", getPresenter().getPersistentState());
    }
    
    @NonNull
    @Override
    public MyPresenter providePresenter() {
        // called when the presenter gets first created and 
        // when the presenter gets created after a process death
        byte[] persistentState = null;
        if (mSavedInstanceState != null) {
            persistentState = mSavedInstanceState.getByteArray("presenterState");
[...]

PR #50, opened 2016-12-19
Provide a generic interface to serialize and deserialize presenters.

vRallev:
This is useful to recreate presenters after a process has died.


vRallev:
I know that you haven't decided what to with #39, yet. But this would be a good compromise in my opinion. If we want to use your library in our app, then we need a solution for this issue.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -52,7 +52,7 @@ public TiPresenter recover(final String id) {
     }
 
     public String safe(@NonNull final TiPresenter presenter) {
-        final String id = generateId(presenter);
+        final String id = presenter.getId();
         TiLog.v(TAG, "safe presenter with id " + id + " " + presenter);
```
passsy: Why is it better to have the presenter generate the id?

vRallev: Because it's the presenter's ID. Otherwise it would be an ID for a presenter used by the `PresenterSavior`. I'm using the ID at other places.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiConfiguration.java
```diff
@@ -198,6 +199,16 @@ public Builder setUseStaticSaviorToRetain(final boolean enabled) {
             return this;
         }
 
+        /**
+         * An {@link TiPresenterSerializer} instance should be provided, if the {@link TiPresenter} should
+         * be restored, after a process has died, but the UI component is being recreated. Since this is
+         * a rare use case the default value is {@code null} and presenters aren't restored.
+         * @param serializer An implementation which can serialize and deserialize any presenter.
+         */
+        public Builder setPresenterSerializer(TiPresenterSerializer serializer) {
```
passsy: Having the serializer as part of the configuration is interesting, I'll have to think about it

vRallev: That was the ideal place for me. It's pretty handy with the global config and optional for tests.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/serialize/TiPresenterSerializer.java
```diff
@@ -0,0 +1,40 @@
...
+
+import net.grandcentrix.thirtyinch.TiPresenter;
+import net.grandcentrix.thirtyinch.TiView;
+
+/**
+ * Interface to serialize and deserialize presenters. The interface should be implemented, if
+ * a {@link TiPresenter} should be restored, after a process has died, but the UI component is
+ * being recreated.
+ */
+public interface TiPresenterSerializer {
```
vRallev: @passsy After reviewing all the changes in the 0.8.0 version I think this interface can be removed and the new `TiPresenterSavior` could be used instead. After that it would be possible for me to release the module in #51 separately and I could finally stop working on the fork.

The only changes that would be necessary to make this work is to add the `TiPresenterSavior` as argument in the `TiConfiguration`. This way I get a chance to replace the default savior. 

The 2nd change that's necessary is to modify the recover method in the savior interface `TiPresenter recover(String presenterId, @NonNull Object host);`. Here a third parameter would be necessary that you've already implemented: `TiPresenterProvider`. In most cases that would be the `TiFragment` or `TiActivity`. That would give me the option to construct a new presenter in case it can be recovered after the process died.

Would that be possible?

vRallev: I think both changes are safe, because nothing of that was exposed to the public API, yet. Do you want me to create a different PR?

vRallev: Dang, I just realized that it's not necessary to make any changes in the config class, because in order to access the config, you need to create a presenter first. 

There's no way to replace the savior in the delegate from `TiFragment` or `TiActivity` at the moment. I think it's better to add a new provider interface for the savior and the default implementation delegate to the `PresenterSavior`. Although a global config would be nicer.

passsy: I'll look into presenter serialisation for the next release and we'll find a way to provide you an API.
The `PresenterSavior` is the new heart of this library. I don't think the savior will be replaceable with a third party impl because small mistakes could lead to major leaks. Instead I'm thinking about APIs for the savior to control the presenter store. Haven't thought this through completely but you get the direction.

passsy:
I still don't get why a new API is required. I implemented saving presenter state with `android-state` in a few seconds. What is missing?

see https://github.com/grandcentrix/ThirtyInch/compare/feature/save_state

This implementation could be moved into a `BasePresenter`/`BaseActivity` in your project. A fork is not required.

vRallev:
This would be tied to the activity lifecycle again. That's not what you want in case a longer operation is going on. Then there wouldn't be a good hook to restore the presenter.

PR #51, opened 2016-12-19
Android State Serializer

vRallev:
This is an implementation for #50. Feel free to close the PR if you don't want to have a dependency to Android State. It's only a reference.


passsy:
Having it as a separate module is fine. But I think it will not be part of this repository.
The default implementation will hopefully work without dependnecies. Depending on how convenient it is to work with this will go in or not. I'll leave this open for reference.

Rainer-Lang:
Maybe it's possible to have this as an extension.

vRallev:
It's an optional extension like the `plugin` or `rx` module.

Rainer-Lang:
Great!

vRallev:
I migrated the branch to our own state saving implementation [Android State](https://github.com/evernote/android-state).

PR #52, opened 2017-01-03
Suppress unchecked casts warnings

vRallev:
It might be minor, but if you compile the project in a terminal, then you see these warnings
```
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```

If you enable the Lint rule, then you can find the classes, e.g.
```
/../ThirtyInch/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/callonmainthread/CallOnMainThreadInterceptor.java:53: warning: [unchecked] unchecked cast
        final V wrappedView = (V) Proxy.newProxyInstance(
                                                        ^
  required: V
  found:    Object
  where V is a type-variable:
    V extends TiView declared in method <V>wrap(V)
```

Problem is that unchecked casts are only suppressed by a comment. IntelliJ recognizes the comment, but not the Java compiler. That's why it's better to suppress those warnings with an annotation.

### DIFF
--- a/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
+++ b/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
@@ -101,12 +101,12 @@ public class TiActivityPlugin<P extends TiPresenter<V>, V extends TiView> extend
         return mDelegate.getPresenter();
     }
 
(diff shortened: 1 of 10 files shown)


vRallev:
Do you want to merge the change?

passsy:
Sad that there is no `SuppressWarnings` for single lines :(

Issue #55, opened 2017-01-06
Behaviours in RxUtils(Test)


StefMa:
While I'm working on #54 I've found some strange stuff.

1. The test `testManageViewSubscription_WithDetachSingleSub_ShouldUnsubscribe` inside `RxTiPresenterSubscriptionHandlerTest` is equal to `testManageViewSubscription_WithDetachView_ShouldUnsubscribe`. So I guess we can remove it?!
Probably I've copied the test code to the `rx2` module which can be removed ❌  as well

2. `testManageViewDisposable_DetachBeforeAttach_ShouldThrowAssertError` in `RxiTiPresenterDisposableHandlerTest` (and maybe in `RxTiPresenterSubscriptionHandlerTest` too) will - hopefully - never happen.
Is it currently possible to detach a view without attaching? 🤔 

3. Should we move the `TiMockPresenter` inside the `test` module? 🤔

StefMa:
Nr 2 is fixed inside #58

StefMa:
Nr 3 is fixed in #58. Because we don't need it in there...

StefMa:
Nr 1 isn't critical.
So I will close this issue :) 👍

PR #54, opened 2017-01-06
RxJava2 support

StefMa:
## RxJava2 support

This is a first draft of supporting RxJava2.
Included: 
The fully helper class `RxTiPresenterDisposableHandler`.
A part of RxTiPresenterUtils. Currently only `isViewReady()`.

I've created a new `rx2` module which can be plugged in like the RxJava (1) module (need to be uploaded to maven of course)...

### TODOs
- [ ]  Add `deliverLatestCacheToView()`
- [ ] Add `deliverLatestToView()`
- [ ] Add `deliverToView()`

PR #58, opened 2017-01-09
Fix manageViewSubcription when there isn't any view attached

StefMa:
I found a strange behaviour during my implementation of #54 (and the follow up issue #55 (2)).
We can actually send a `viewSubscription` to the SubscriptionHandler even when there isn't any view attached.

This PR will fix that issue.


passsy:
I'm fine with throwing an exception when calling `manage*()` and the presenter is in the wrong state

PR #64, opened 2017-01-24
Use addAll instead of our own fori

### LINKED ISSUE #63: User `addAll` instead of using an fori
Currently we are using an `fori` loop to add subscriptions to the CompositeSubscprion handler.
See [here](https://github.com/grandcentrix/ThirtyInch/blob/master/rx/src/main/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandler.java#L96).

I found out that the `CompositeSubscrption` (and the RxJava2 part) have a `addAll(...)` method.
Think about to use these one instead of our own `fori` loop

StefMa:
This will fix #63

### DIFF
--- a/rx/src/main/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandler.java
+++ b/rx/src/main/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandler.java
@@ -68,7 +68,7 @@ public class RxTiPresenterSubscriptionHandler {
                     + " when the presenter has reached the DESTROYED state");
         }
 
-        addSubscriptions(mPresenterSubscriptions, subscriptions);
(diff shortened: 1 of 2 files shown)


StefMa:
`CompositeSubscription` and `CompositeDisposable` have a different behavior in `addAll`.
See [this issue on RxJava](https://github.com/ReactiveX/RxJava/issues/5016).

I've already replace or own `fori` with `addAll`, even in rx2.
Should we "revert" it and use our previous implementation?! 🤔

passsy:
I prefer `addAll` because `CompositeSubscription` and `CompositeDisposable` handle concurrency better than our `fori` solution. 

The difference between `rx1` and `rx2` is interesting but not a big deal. In the end every `Disposable` will be freed once `dispose()` gets called in `TiPresenter#onDestroy()`

PR #65, opened 2017-01-27
TiPresenter#runOnUiThread(Runnable)

passsy:
Allows code to run on the UI thread within the `TiPresenter`. This is most likely **not** a method for users of `Ti` but for extensions using the `LifecycleObserver`-API (another PR will follow).

### public API changes
The action of `TiPresenter#sendToView(Runnable action)` will be now be called on the UI thread. Furthermore will those actions be executed after `onAttachView(TiView)` and after all lifecycle observers have been called. This allows preparing the view in `onAttachView(TiView)` for those actions. That way the view should be in a "running" state as if the view was never gone.
For very rare and special cases `TiPresenter#getQueuedViewActions()` can be used to manually execute queued actions. People are creative.

Another change has been made to `RxTiPresenterUtils#isViewReady()` which was firing the ready event before `onAttachView(TiView)` was called. The ready event is now fired after `onAttachView(TiView)`

The execution order of lifecycle obersvers when receiving teardown events has been reversed; following "first in, last out"

### internal changes

The method in `DelegatedTiActivity#postToMessageQueue(Runnable runnable)` has been replaced with `DelegatedTiActivity#Executor getUiThreadExecutor();`, same for the `Fragment` interface. All classes have been adjusted.

---

I think the API changes are not critical and could also be bug fixes preventing rare race conditions.

### DIFF
--- a/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
+++ b/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
@@ -29,6 +29,7 @@ import net.grandcentrix.thirtyinch.internal.TiActivityDelegate;
 import net.grandcentrix.thirtyinch.internal.TiLoggingTagProvider;
 import net.grandcentrix.thirtyinch.internal.TiPresenterProvider;
 import net.grandcentrix.thirtyinch.internal.TiViewProvider;
+import net.grandcentrix.thirtyinch.internal.UiThreadExecutor;
 import net.grandcentrix.thirtyinch.util.AndroidDeveloperOptions;
 import net.grandcentrix.thirtyinch.util.AnnotationUtil;
 
@@ -39,6 +40,7 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import java.util.List;
+import java.util.concurrent.Executor;
 
 /**
  * Binds a {@link TiPresenter} to an {@link Activity}
@@ -118,6 +120,11 @@ public class TiActivityPlugin<P extends TiPresenter<V>, V extends TiView> extend
         return null;
     }
 
+    @Override
+    public Executor getUiThreadExecutor() {
+        return new UiThreadExecutor();
+    }
+
     /**
      * Invalidates the cache of the latest bound view. Forces the next binding of the view to run
      * through all the interceptors (again).
@@ -192,11 +199,6 @@ public class TiActivityPlugin<P extends TiPresenter<V>, V extends TiView> extend
         mDelegate.onStop_beforeSuper();
         super.onStop();
         mDelegate.onStop_afterSuper();
-    }
-
-    @Override
-    public boolean postToMessageQueue(final Runnable runnable) {
-        return getActivity().getWindow().getDecorView().post(runnable);
     }
 
     @SuppressWarnings("unchecked")
--- a/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
+++ b/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
@@ -59,6 +59,8 @@ public class TiActivityPlugin<P extends TiPresenter<V>, V extends TiView> extend
 
     private TiActivityDelegate<P, V> mDelegate;
 
+    private final UiThreadExecutor mUiThreadExecutor = new UiThreadExecutor();
+
     /**
      * Binds a {@link TiPresenter} returned by the {@link TiPresenterProvider} to the {@link
      * Activity} and all future {@link Activity} instances created due to configuration changes.
@@ -122,7 +124,7 @@ public class TiActivityPlugin<P extends TiPresenter<V>, V extends TiView> extend
 
     @Override
     public Executor getUiThreadExecutor() {
-        return new UiThreadExecutor();
+        return mUiThreadExecutor;
     }
 
     /**
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiActivity.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiActivity.java
@@ -50,6 +50,8 @@ public abstract class TiActivity<P extends TiPresenter<V>, V extends TiView>
     private final TiActivityDelegate<P, V> mDelegate
             = new TiActivityDelegate<>(this, this, this, this);
 
+    private final UiThreadExecutor mUiThreadExecutor = new UiThreadExecutor();
+
     @NonNull
     @Override
     public Removable addBindViewInterceptor(@NonNull final BindViewInterceptor interceptor) {
@@ -94,7 +96,7 @@ public abstract class TiActivity<P extends TiPresenter<V>, V extends TiView>
 
     @Override
     public Executor getUiThreadExecutor() {
-        return new UiThreadExecutor();
+        return mUiThreadExecutor;
     }
 
     /**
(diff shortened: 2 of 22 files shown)


thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiFragment.java
```diff
@@ -73,6 +75,11 @@ public P getPresenter() {
         return mDelegate.getPresenter();
     }
 
+    @Override
+    public Executor getUiThreadExecutor() {
+        return new UiThreadExecutor();
```
albrecht87: ohne genau den Kontext zu kennen: Soll hier immer ein neuer erstellt werden, oder ist die Anzahl begrenzt?

passsy: It's harder to leak an Activity when the `Executor` isn't `static` or reused. Also it's a very small class, shouldn't cause problems

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
```diff
@@ -315,6 +324,32 @@ public boolean isViewAttached() {
         return mState == State.VIEW_ATTACHED;
     }
 
+    /**
+     * Runs the specified action on the UI thread. It only works when a view is attached
+     * <p>
+     * When you are looking for a way to execute code when the view is attached have a look
```
StefMa: > When you are looking for a way to execute code when the view **got available sometime in the future** [...]

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
```diff
@@ -315,6 +324,32 @@ public boolean isViewAttached() {
...
+     * at {@link #sendToView(ViewAction)}
+     *
+     * @param action the action to run on the UI thread
+     * @throws IllegalStateException when the executor is not available
+     */
+    public void runOnUiThread(@NonNull final Runnable action) {
```
StefMa: Please add some tests for this method.
Also when it is not really usable for "users"... But it is `public`. So we need to take care of it.

Also thinking about to make it just `protected`... Isn't it enough? 🤔

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/UiThreadExecutorAutoBinder.java
```diff
@@ -0,0 +1,51 @@
...
+ * binds a ui thread executor to the presenter when this view is attached
+ */
+public class UiThreadExecutorAutoBinder implements TiLifecycleObserver {
+
+    private final TiPresenter mPresenter;
+
+    private final Executor mUiThreadExecutor;
+
+    public UiThreadExecutorAutoBinder(final TiPresenter presenter,
+            final Executor uiThreadExecutor) {
```
StefMa: When we expect a `UiThreadExecutor`, then change the `Executor` signature to it.
Otherwise we can put `any` Executor to it which isn't on the UiThread.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/DelegatedTiActivity.java
```diff
@@ -33,6 +35,11 @@
     P getRetainedPresenter();
 
     /**
+     * @return {@link UiThreadExecutor}
+     */
+    Executor getUiThreadExecutor();
```
StefMa: This is the same like the `AutoBinder`. Change it to UiTheadExecutor as return value...

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/DelegatedTiFragment.java
```diff
@@ -18,9 +18,16 @@
...
 import android.support.v4.app.Fragment;
 
+import java.util.concurrent.Executor;
+
 public interface DelegatedTiFragment {
 
     /**
+     * @return {@link UiThreadExecutor}
+     */
+    Executor getUiThreadExecutor();
```
StefMa: Same here :)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiActivity.java
```diff
@@ -90,6 +92,11 @@ public P getRetainedPresenter() {
         return null;
     }
 
+    @Override
+    public Executor getUiThreadExecutor() {
```
StefMa: Same here

plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
```diff
@@ -118,6 +120,11 @@ public P getRetainedPresenter() {
         return null;
     }
 
+    @Override
+    public Executor getUiThreadExecutor() {
```
StefMa: Same here :)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiDialogFragment.java
```diff
@@ -74,6 +76,11 @@ public P getPresenter() {
         return mDelegate.getPresenter();
     }
 
+    @Override
+    public Executor getUiThreadExecutor() {
```
StefMa: Same here

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiFragment.java
StefMa: Same here :)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
passsy: sure, first of all I want an overall acceptance before I invest time to test his approach

passsy: cannot be `protected` because the View implementer (Activity via delegate) has to set it, therefore it has to be `public` because they live in a different package.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/UiThreadExecutorAutoBinder.java
passsy: This would break our compatibility with JVM tests. The delegates and all used interfaces are android classes free. The `UiThreadExecutor` is 100% Android related.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/DelegatedTiActivity.java
passsy: breaks JVM tests

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
```diff
@@ -506,8 +552,20 @@ private void moveToState(final State newState, final boolean hasLifecycleMethodB
...
+                    mLifecycleObservers.get(i).onChange(newState, hasLifecycleMethodBeenCalled);
+                }
+                break;
+
+            case VIEW_DETACHED:
+            case DESTROYED:
+                // reverse observer order for teardown events; first in, last out
+                for (int i = mLifecycleObservers.size() - 1; i >= 0; i--) {
+                    mLifecycleObservers.get(i).onChange(newState, hasLifecycleMethodBeenCalled);
+                }
```
passsy: - [ ] add tests for correct call oder

passsy:
@StefMa Ready. I already tested the sample and a real project and everything works fine

PR #69, opened 2017-02-27
Add retrolambda to the sample app

jonathan-caryl:
Adding retrolambda makes for more compact and easier to read code when
used with RxJava

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -1,3 +1,7 @@
+plugins {
+    id "me.tatarka.retrolambda" version "3.5.0"
+}
+
 apply plugin: 'com.android.application'
 
 android {
@@ -19,6 +23,11 @@ android {
     }
     lintOptions {
         abortOnError false
+    }
+
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
     }
 }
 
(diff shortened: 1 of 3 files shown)


StefMa:
Don't know if I like that Idea 🤔
Samples are here to understand the simplest use case of an implementation.
Unfortunately not all developers are similar with lambdas yet (and but
true, I know).

However. If we add something like that I recommend to add it with the
official implementation of android (have never used it,so don't know if it
will work with our sample).

Anyway what do you say @passsy?

jonathan-caryl:
You could say that not all developers are familiar with RxJava yet either...

passsy:
I'm fine with using lambdas in the sample. However, we should not integrate retrolambda and use the jack compiler instead.

jonathan-caryl:
I've not used the jack compiler yet. I'll see if I can figure that out & update the PR.

jonathan-caryl:
Looks like using Jack is now pretty simple. PR updated.

vRallev:
Note that Jack is much slower than RetroLamba. But that's probably not a big deal for this sample.

StefMa:
I've noticed it as well @vRallev .
But as you said. It is an sample ;)

If I clean the project and run it it cost me round about 35 secs.
If I change something in the code it builds in round about 2 secs.

I guess that is enough.

Thank you @jonathan-caryl for the PR 👍

passsy:
travis hangs at `:sample:transformClassesWithPreJackRuntimeLibrariesForDebug` 😒

jonathan-caryl:
[SO](http://stackoverflow.com/a/41664491/1789683) suggests this may be solved by bumping up the heap size for dex-ing. Let's give that a go...

jonathan-caryl:
Back to retrolambda

StefMa:
You have merged a broken build 🤣

passsy:
travis was fine, only codecov wasn't happy

Issue #30, opened 2016-10-30
How to handle onNewIntent()


dave08:
I got a null view in onWakeUp(), if so, where can I yes be assured to have a view ready, isn't that the whole concept of having onWakeUp()?

dave08:
My fault, I accidently called the presenter to update what it's showing in onCreate, since I need to pass it the current object that needs to be displayed from the intent the activity received. Is there any good way to do this? there is nothing in the presenter code that receives intent from onCreate and from onNewIntent, but this is certainly needed to be dealt with by the presenter. I am using RxJava, should I use a BehaviourSubject?

Thanks.

passsy:
The `onActivityResult` case is not really solved as of now. The problem is that `onActivityResult` is called *before* `onCreate`. Saving the result in a variable and request it from the view in `onWakeUp()` is the current "solution".

dave08:
I think I didn't explain myself enough. I am implementing a BrowserActivity with the android WebView. In onCreate I receive an url from intent.getData() and I need to pass it to the SitePresenter to instruct the view to display the site (if allowed by business logic). If a new intent is sent while the activity is open, it should update the current site being displayed by calling onSiteLoadRequest() in the presenter. So, who is responsible in saving the state in onSaveInstanceState and restoring it in onCreate?

Also, I want the WebView to restore itself to the current page when the screen orientation changes, but when I save it's state in onSaveInstanceState and load it in onCreate, my onWakeUp gets called and loads the home page of that site after the web view restores it's own state.

I'm a bit confused what to do in such a situation?

passsy:
ok, simple. Don't use onSaveInstanceState, save the state in the presenter. when you receive a new intent pass it immediately to the presenter (I hope this happens after `Activity#onCreate`), maintain the state there and when the view attaches update it accordingly

dave08:
My problem is that the WebView gives a bundle of its current state, and restores from that bundle. Should I be saving the bundle in the presenter in onDetachView and restoring it at onAttachView, or is this making the presenter too attached to Android api and complicates testing.

passsy:
Oh, interesting. Hmm...
You cannot use `Bundle` in your tests, therefore it's too close to the Android SDK. Sadly I don't have a good solution for you here. WebView Activities often are very simple and don't have business logic. Is MVP here really helpful? 
A WebView is so close to the Android SDK, I don't think a MVP abstraction is very useful here.

dave08:
[...]

PR #70, opened 2017-03-02
Add sendToPresenter(p -> {})  analog to sendToView(v -> {})

passsy:
Like `sendToView(v -> {})` the new `sendToPresenter(p -> {})` is very useful for permissions and activity results. It can be used in callback methods with can be called **before** `#onCreate(Bundle)`

```java

    @Override
    protected void onActivityResult(final int requestCode, final int resultCode,
            final Intent data) {
        super.onActivityResult(requestCode, resultCode, data);
        sendToPresenter(presenter -> presenter.onResultReceived());
    }
```

issues #30 

I also created an interface `PresenterAccessor` bundling `getPresenter()` and `sendToPresenter { }`. Now all methods in Activities and Fragment implementations use `@Override` everywhere. And we don't miss something

### DIFF
--- a/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/FragmentSendToPresenterTest.java
+++ b/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/FragmentSendToPresenterTest.java
@@ -16,8 +16,6 @@
 package net.grandcentrix.thirtyinch;
 
 
-import net.grandcentrix.thirtyinch.internal.TiActivityDelegate;
-import net.grandcentrix.thirtyinch.internal.TiActivityDelegateBuilder;
 import net.grandcentrix.thirtyinch.internal.TiFragmentDelegate;
 import net.grandcentrix.thirtyinch.internal.TiFragmentDelegateBuilder;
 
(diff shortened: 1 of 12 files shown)


thirtyinch/src/test/java/net/grandcentrix/thirtyinch/FragmentSendToPresenterTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+        // will be executed
+        verify(presenter, times(1)).doSomething1();
+    }
+
+    @Test
+    public void testSendToPresenter_BeforePresenterIsAttached() throws Exception {
+
+        final MyPresenter presenter = Mockito.spy(new MyPresenter());
+        final TiActivityDelegate<TiPresenter<TiView>, TiView> delegate
```
StefMa: Same here ©️

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/FragmentSendToPresenterTest.java
```diff
@@ -0,0 +1,159 @@
...
+        void doSomething3() {
+
+        }
+    }
+
+    @Test
+    public void testSendToPresenterInOrder() throws Exception {
+
+        final MyPresenter presenter = Mockito.spy(new MyPresenter());
+        final TiActivityDelegate<TiPresenter<TiView>, TiView> delegate
```
StefMa: Copy and paste error 😉

passsy:
My fault. `onActivityResult` is *always called after* `onCreate` this PR doesn't make any sense. 

Anyways, this `PresenterAccessor` interface was a good idea, I'll extract this part

PR #73, opened 2017-03-06
Manage*Subcription return Subscription now

### LINKED ISSUE #59: Let manageSubscription(Subscription) return Subscription
`RxTiPresenterSubscriptionHandler.manageSubscription(Subscription...)` returns void. It would be cool if there is a `manageSubscription` function with the following signature:

```java 
public Subscription manageViewSubscription(@NonNull final Subscription subscription)
```

Usage
```kotlin
// inside TiPresenter

val rxHelper: RxTiPresenterSubscriptionHandler(this)
val sub: Subscription? = null

/**
 * download some data
 */
fun download() {
    // subscription is managed, it will be cancelled when the Presenter gets destroyed
    sub = rxHelper.manageSubscription(myService.downloadSomething().subscribe{ data -> 
        sendToView { view -> view.showData(data) }
    })
}

/**
 * cancel download before it completes
 */
fun cancelDownload() {
    sub?.unsubscribe()
}
```

StefMa:
This will fix #59 

Nothing special here.
`manageViewSubscription/disposable` will now return the given `Subscription/Disposable`.
`manageViewSubscriptions` have introduced to add multiple `Subscription/Disposable`.

That is a small "API change". But only on RC's.
Because in earlier 0.8 RC's we have changed the API from `manageViewSubscription(Subscription)` to `manageViewSubscription(Subscription...)`.
Not we "revertet" it to `manageViewSubscription(Subscription)` but added `manageViewSubscriptions(Subscription...)` (note the `S` here 😉)

### DIFF
--- a/sample/src/main/java/net/grandcentrix/thirtyinch/sample/HelloWorldPresenter.java
+++ b/sample/src/main/java/net/grandcentrix/thirtyinch/sample/HelloWorldPresenter.java
@@ -24,6 +24,7 @@ import android.support.annotation.NonNull;
 import java.util.concurrent.TimeUnit;
 
 import rx.Observable;
+import rx.Subscription;
 import rx.functions.Func1;
 import rx.schedulers.Schedulers;
 import rx.subjects.BehaviorSubject;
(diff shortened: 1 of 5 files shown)


sample/src/main/java/net/grandcentrix/thirtyinch/sample/HelloWorldPresenter.java
```diff
@@ -44,13 +45,13 @@
...
-
-        rxSubscriptionHelper.manageViewSubscription(view.onButtonClicked()
+        final Subscription showTextSub = mText.asObservable().subscribe(view::showText);
+        final Subscription onButtonClickSub = view.onButtonClicked()
                 .subscribe(aVoid -> {
                     triggerHeavyCalculation.onNext(null);
-                }));
+                });
+
+        rxSubscriptionHelper.manageViewSubscriptions(showTextSub, onButtonClickSub);
```
passsy: I'm not sure if it's useful to create local variables for the subscriptions. But I also don't like method references. Just personal preference here 🦄

PR #75, opened 2017-03-09
Make getPresenter an interface so we never forget it

passsy:
The good parts of #70


passsy:
`PresenterNonConfigurationInstance`is something completely different and shouldn't implement `PresenterAccessor`.

PR #77, opened 2017-03-15
Single support

### LINKED ISSUE #40: rx.Single support for RxTiPresenterUtils
Most of my data is returned as Singles instead of Observables, so having Single support built-in for RxTiPresenterUtils would be great.

I can of course work around this limitation by simply doing this:
```java
categoryService.getMainCategories()
    .toObservable()
    .compose(RxTiPresenterUtils.deliverLatestToView(this))
    .subscribe(categories -> ()));
```

StefMa:
This will fix #40 

Because in Java it is not possible to have the same Signature but a different return value I've splittet the RxTiPresenterUtils into three parts:
* `RxTiObservableUtils`
* `RxTiUtils`
* `RxTiSingleUtils`
`RxTiPresenterUtils` is deprecated now. 

I've removed the `Presenter` in the class name because you can use it in Activtiys as well..
It is really not recommended but not mandatory...

However. I've also changed these "pattern" in Rx2.
But at this time, because we have no `deliver*View` methods, only the `RxTiUtils` class exist.
We need in Rx2 also a `RxTiMaybeUtils` :)
But that will be done in the future...

I've also added a simple example in our sample.

### DIFF
--- a/rx/src/main/java/net/grandcentrix/thirtyinch/rx/RxTiObservableUtils.java
+++ b/rx/src/main/java/net/grandcentrix/thirtyinch/rx/RxTiObservableUtils.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright (C) 2017 grandcentrix GmbH
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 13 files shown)


StefMa:
Because that solution is not acceptable (we have talked about that internally already) I will close that for now...

Issue #33, opened 2016-11-02
Presenter not destroyed when Fragment is popped off the backstack


gborland:
When the back key is pressed to pop a `TiFragment` off the backstack, the Presenter does not get destroyed. Here is the log output from `TiFragment`:
```
11-02 09:17:04.985 V: onStop()
11-02 09:17:04.985 V: onDestroyView()
11-02 09:17:04.986 V: onDestroy() recreating=true
11-02 09:17:04.986 V: onDetach()
```

Because the activity is not being destroyed, the `destroyPresenter` flag in `TiFragment#onDestroy()` does not get set to true. This means that the Presenter is leaked.

gborland:
I don't really understand why `TiFragment#onDestroy()` has all that logic to decide whether or not to destroy the Presenter.

If `config.shouldRetainPresenter()` is true, which is the default, then the fragment has `setRetainInstance(true)` set in its `onCreate()` anyway, so the Fragment won't get destroyed on a normal configuration change. If `Fragment.onDestroy()` is being called by the framework, it's because the Fragment really is being properly destroyed, so we should destroy the presenter too.

passsy:
Good catch, will investigate it

passsy:
I can verify this problem exists. But the fix isn't as easy always calling `Presenter.destroy()` in `TiFragment#onDestroy()`. The following states must be tested:

- with/without savior
- with/without retain
- don't keep activities enabled/disabled
- fragment in backstack
- fragment attached/detached

for the following usecases:
- call `finish()`
- bring Activity in background and back to foreground
- orientation change

Overall 72 cases, which should be tested. It was easier for the Activity where only 8 states are possible for 3 usecases (24 test). One reason I don't use Fragments anymore 😈 
This also requires a delegate for `TiFragment`, otherwise it's not testable.
//cc #20

passsy:
Interestingly, there is a bugfix in supportLibrary `25.1.0` 
 - `Fragment.onDestroy()` not called for fragment in backstack

Maybe this caused some of the troubles

rberghegger:
@passy `onDestroy()` is reliably called each time a fragment gets popped off the back stack. Even in support library version `24.2.1` which is currently used by ThirtyInch.
Sadly the bugfix description does not provide any helpful informations (for example under what conditions `onDestroy()` was not called) :(. Nevertheless the memory leak which is described in this issue can be verified and needs to be fixed.

passy:
[...]

PR #78, opened 2017-03-20
Bugfix: Presenter not destroyed when Fragment is popped off the backstack

rberghegger:
Hello everyone,

currently I'm working on a fix for #33.
We thought it would be better if we add all relevant Unit-Tests first so it would be easier to develop the bugfix. Additionally I've created a new Activity for the "Sample" application which can be be used to interactively test the Fragment lifecycle under different circumstances (e.g. back stack or configuration change).

I would like to open this PR as early as possible to make it open for discussion.

Please feel free to comment and contribute!


passsy:
I added stubs for all testcases and moved the current implementations to the matching method. There have been already two different implementations for case `saviorTrue_retainTrue_backstackFalse_dontKeepActivitiesTrue_activityFinishing` previously called `saviorTrue_retainTrue_dontKeepActivitiesTrue_activityFinishing` and `saviorTrue_retainTrue_dontKeepActivitiesTrue_fragmentAttached_activityFinishing`. I wasn't sure what "fragmentAttached" meant.

<details>
<summary>
Method generator code, in case we have to adjust the method names in the future
</summary>
fun main(args: Array<String>) {

    for (savior in listOf("true", "false")) {
        for (retain in listOf("true", "false")) {
            for (dontKeepActivities in listOf("true", "false")) {
                for (backstack in listOf("true", "false")) {
                    for (action in listOf("activityFinishing", "activityChangingConfiguration",
                            "popBackstack")) {

                        if (backstack == "false" && action == "popBackstack") {
                            // not a valid case
                            continue
                        }

                        val impl = """
                        |        final TestPresenter presenter = new TestPresenter(new TiConfiguration.Builder()
                        |                  .setUseStaticSaviorToRetain($savior)
                        |                  .setRetainPresenterEnabled($retain)
                        |                  .build());
                        |
                        |          final TiFragmentDelegate<TiPresenter<TiView>, TiView> delegate
                        |                  = new TiFragmentDelegateBuilder()
                        |                  .setDontKeepActivitiesEnabled($dontKeepActivities)
                        |                  .setIsAdded(true)
                        |                  .setIsHostingActivityFinishing(${if(action == "activityFinishing") "true" else "false"})
                        |                  .setIsHostingActivityChangingConfiguration(${if(action == "activityChangingConfiguration") "true" else "false"})
                        |                  .setSavior(mSavior)
                        |                  .setPresenter(presenter)
                        |                  .build();
                        |
                        |          // TODO implement testcase
                        |          fail();
                        """.replaceIndentByMargin()


                        println("public void savior${savior.capitalize()}_" +
                                "retain${retain.capitalize()}_" +
                                "backstack${backstack.capitalize()}_" +
                                "dontKeepActivities${dontKeepActivities.capitalize()}_" +
                                "$action(){\n$impl\n}")
                    }
                }
            }
        }
    }
}

</details>

rberghegger:
Great! 👍  I guess `fragmentAttached` was a fallacy :-D

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/TiFragmentPresenterDestroyTest.java
```diff
@@ -176,8 +168,35 @@ public void saviorFalse_retainFalse_backstackFalse_dontKeepActivitiesTrue_activi
...
+        delegate.onCreate_afterSuper(null);
+        delegate.onCreateView_beforeSuper(mock(LayoutInflater.class), null, null);
+        delegate.onStart_afterSuper();
+
+        // And when the Activity is changing configurations.
+        delegate.onSaveInstanceState_afterSuper(mSavedState);
+        delegate.onStop_beforeSuper();
+        delegate.onDestroyView_beforeSuper();
+        delegate.onDestroy_afterSuper();
+
```
passsy: add assertions here:
```diff
+        // Then assert that the Presenter is destroyed and not saved in the savior.
+        assertThat(delegate.getPresenter().isDestroyed()).isTrue();
+        assertThat(mSavior.presenterCount()).isEqualTo(0);
```

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/TiFragmentPresenterDestroyTest.java
```diff
@@ -278,17 +285,43 @@ public void saviorFalse_retainFalse_backstackTrue_dontKeepActivitiesFalse_activi
...
+        // When a new Activity gets recreated
+        final HostingActivity hostingActivity2 = new HostingActivity();
+
+        // And a new Fragment instance is created by the framework.
+        final TestPresenter presenter2 = new TestPresenter(new TiConfiguration.Builder()
+                .setUseStaticSaviorToRetain(false)
+                .setRetainPresenterEnabled(false)
+                .build());
+        final TiFragmentDelegate<TiPresenter<TiView>, TiView> delegate2
```
rberghegger: This test does also test what happens if the back stack is popped after the Activity was recreated after a configuration change. So either the name of the test must be changed or we remove the additional code. What do you say, @passsy?

passsy:
I think I solved it!

First of all I made a few decisions:

1.  _Presenters will be destroyed when the Fragment is not managed by the FragmentManager._ This happens when FragmentManager#remove or replaced is called and the `Fragment` is not in the Backstack.
The Presenter works now like the `savedInstanceState` but of course it doesn't survive process death.

2. _TiConfiguration#setUseStaticSaviorToRetain(Boolean) will not be supported in the future_. First of all nobody uses it, second it's a implementation detail of this library and especially when using `Fragments` there is no way to make `Presenters` survive without the `Savior`.

3. _TiFragment will not work with `setRetainInstanceState(true)`._ I've learned that only Fragments without UI should set `retainInstanceState` to `true`. Since `setUseStaticSaviorToRetain(false)` was deprecated we can always use the `Savior` and everything works fine.

### Solution

The solution requires the `ActivityLifecycleCallbacks` to detect when `Activities` get destroyed. Here is a Usecase which makes a lot of trouble:

- Add `FragmentA` with backstack `true`
- Replace `FragmentA` with another Fragment (B). `FragmentA` is not detached in the backstack, only `onDestroyView` got called, not `onDestroy`
- Trigger Activity configuration change. `FragmentA#onDestroy()` and`FragmentA#onDetach()` got called. The `FragmentManager` only holds a serialised `FragmentState` from `FragmentA`, no instance exists.
- Finish Activity. Because no instance of `FragmentA` exists the `FragmentManager` clears the `FragmentState` without any callback. Normally the `TiPresenter` of `FragmentA` would leak here.
- New: `ActivityLifecycleCallbacks` detect the finished `Activity` and all associated `Presenter` and destroys them properly.

### Missing

- [ ] Using `ActivityLifecycleCallbacks` makes testing currently impossible
- [ ] Documentation about deprecation of `TiConfiguration#setUseStaticSaviorToRetain(Boolean)`
- [ ] Prevent setting `TiFragment#setRetainInstanceState(true)`. 

### Please verify before merge:
- [ ] same Fragment instance added and removed. Now added again. A new presenter instance should be created.
- [ ] How does the Fragments behave in a `ViewPager`

ar-g:
@passsy I didn't understand the use of `ActivityLifecycleCallbacks`.
I thought this check should fix backstack problem
```java
if (!mTiFragment.isInBackstack()) 
```

passsy:
@ar-g the problem is that there is no `Fragment` instance where this code could be executed. When a `Fragment` is in backstack and the `Activity` get's recreated the `FragmentManager` only holds a `FragmentState` containing all information about the `Fragment`. Just a simple record. At this time the `Presenter` is still alive because a `Fragment` instance could be instantiated from the `FragmentManager` by popping the backstack.
 
When the `Activity` gets destroyed the `FragmentManager` drops this record. It doesn't create a `Fragment` instance which could give any callback. Because there is no callback I can't destroy the `Presenter` and it will leak. That's why I listen for the `Activity` to finish and destroy all remaining `Presenters` (those for `FragmentState` backstack entries) associated with this Activity.

passsy:
The solution is ready all tests are finally green. Some documentation is missing and a lot of testing!

gborland:
I'd be happy to test this. Is there a snapshot build I can grab with gradle?

passsy:
we don't have public snapshots, sorry. But you can build the project and publish to your local maven.

```
// publish aars to mavenLocal()
./gradlew clean bundleRelease publishToMavenLocal`
```

add this to your project Top-level `build.gradle`:
```
allprojects {
    repositories {
        jcenter()
        mavenLocal() // access your local maven
    }
}
```

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ActivityScopedPresenters {
+
+    private final String TAG = ActivityScopedPresenters.class.getSimpleName() + "@" + Integer
+            .toHexString(hashCode());
+
+    private final HashMap<String, TiPresenter> store = new HashMap<>();
```
StefMa: mStore

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
```diff
@@ -0,0 +1,58 @@
...
+import net.grandcentrix.thirtyinch.TiLog;
+import net.grandcentrix.thirtyinch.TiPresenter;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ActivityScopedPresenters {
```
StefMa: Don't know if `ActivtiyScopedPresenters` a good name 🤔 
As far is I see it right, the class holds a Map of Strings and Presenters... Not more.
So, why `Activity` inside the name?
And why `scoped`? 
I think about an `TiPresenterHolder` or smth...

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
 
-    INSTANCE;
+    private static PresenterSavior INSTANCE;
 
     private static final String TAG = PresenterSavior.class.getSimpleName();
 
-    private final HashMap<String, TiPresenter> mPresenters = new HashMap<>();
+    static final String TI_ACTIVITY_PRESENTER_SCOPE_KEY = "ThirtyInch_presenter_scope_id";
+
+    @VisibleForTesting
```
StefMa: VisibleForTesting brings an `otherwise` parameter... It would be great to use it 👍

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+    Application.ActivityLifecycleCallbacks mLifecycleCallbacks;
+
+    @VisibleForTesting
```
StefMa: VisibleForTesting brings an `otherwise` parameter... It would be great to use it 👍

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+    private final HashMap<Activity, String> mScopeIdForActivity = new HashMap<>();
+
+    public static synchronized PresenterSavior getInstance() {
+        if (INSTANCE == null) {
+            INSTANCE = new PresenterSavior();
+        }
+        return INSTANCE;
+    }
+
+    @VisibleForTesting
```
StefMa: Same here :)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+    /*package*/ PresenterSavior() {
```
StefMa: I haven't ever seen smth like that before in this project.. So remove it please :)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+
     private String generateId(@NonNull final TiPresenter presenter) {
         return presenter.getClass().getSimpleName()
                 + ":" + presenter.hashCode()
                 + ":" + System.nanoTime();
     }
+
+    @NonNull
+    private Application.ActivityLifecycleCallbacks getActivityLifecycleCallbacks() {
+        return new Application.ActivityLifecycleCallbacks() {
```
StefMa: Please extract this class...

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
-    private final HashMap<String, TiPresenter> mPresenters = new HashMap<>();
+    static final String TI_ACTIVITY_PRESENTER_SCOPE_KEY = "ThirtyInch_presenter_scope_id";
+
+    @VisibleForTesting
+    Application.ActivityLifecycleCallbacks mLifecycleCallbacks;
+
+    @VisibleForTesting
+    final HashMap<String, ActivityScopedPresenters> mScopes = new HashMap<>();
+
+    private final HashMap<Activity, String> mScopeIdForActivity = new HashMap<>();
```
StefMa: I don't know if "scope" is the right wording here. 
The String is more a "activity identifier" than a scope 🤔 , right?

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+                mScopes.put(scopeId, scope);
+            } else {
+                scope = gotScope;
+            }
+        }
+
+        return scope;
+    }
+
+    private void printRemainingStore() {
```
StefMa: This is only for debugging right? 🤔 
I would prefer to add these info to the method name.
I also recomment to create a static boolean which can be toggled manually to identify if we have to run this code or not...

```
privat static final boolean DEBUG = true;

private void printDebugRemainingStore() {
    if(DEBUG) {
        // Run that code
    }
}
```

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+                }
+                scope.remove(key);
+            }
+        }
+
+        printRemainingStore();
+    }
+
+    @VisibleForTesting
+    /*package*/ int getPresenterCount() {
```
StefMa: Remove /*package*/

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+                    presenter.destroy();
+                }
+                scope.remove(key);
+            }
+        }
+
+        printRemainingStore();
+    }
+
+    @VisibleForTesting
```
StefMa: Use otherwise

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
     private static final String TAG = PresenterSavior.class.getSimpleName();
 
-    private final HashMap<String, TiPresenter> mPresenters = new HashMap<>();
+    static final String TI_ACTIVITY_PRESENTER_SCOPE_KEY = "ThirtyInch_presenter_scope_id";
+
+    @VisibleForTesting
+    Application.ActivityLifecycleCallbacks mLifecycleCallbacks;
+
+    @VisibleForTesting
+    final HashMap<String, ActivityScopedPresenters> mScopes = new HashMap<>();
```
StefMa: `mPresenterIdForPresenter` 🤔

dave08:
Look into jetpack.io (i think thats the name... many projects use this), it
gives you gradle dependency for any commit in a github repo. It also tells
you on its ui if the compilation was successful when you try resolving your
dependencies. Good luck☺

בתאריך 19 באפר׳ 2017 07:44,‏ "Pascal Welsch" <notifications@github.com> כתב:

StefMa:
Jep. Like @dave08 said. You can use jitpack.io
-> https://jitpack.io/#grandcentrix/thirtyinch/bugfix~fragment_backstack_presenter_not_destroyed-SNAPSHOT

I've successfully triggered the build. So that this branch is ready to use ;)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+        if (scope == null) {
+            return null;
+        }
+        return scope.get(presenterId);
     }
 
     @Override
-    public String save(@NonNull final TiPresenter presenter) {
+    public String save(@NonNull final TiPresenter presenter, @NonNull final Activity activity) {
         final String id = generateId(presenter);
```
StefMa: rename `id` to `presenterId`

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+
+        final ArrayList<TiPresenter> presenters = new ArrayList<>();
+        for (final Map.Entry<String, ActivityScopedPresenters> entry : mScopes.entrySet()) {
+            presenters.addAll(entry.getValue().getAll());
+        }
+        return presenters.size();
+    }
+
+    private synchronized void clearScopeWhenPossible(@NonNull final Activity activity) {
+        final String scopeId = mScopeIdForActivity.get(activity);
```
StefMa: As I already said (feels like 100 times :D) scope is not a good naming here.
Thinking about `activtiyId` 🤔

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/TiActivityDelegate.java
```diff
@@ -249,7 +256,8 @@ public void onDestroy_afterSuper() {
 
         if (destroyPresenter) {
             mPresenter.destroy();
-            mSavior.free(mPresenterId);
+
```
StefMa: Remove that new line.. I don't increase the readability...

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/TiFragmentDelegate.java
```diff
@@ -225,10 +253,11 @@ public void onDestroy_afterSuper() {
 
         if (destroyPresenter) {
             mPresenter.destroy();
-            mSavior.free(mPresenterId);
+
```
StefMa: Remove that new line

thirtyinch/src/androidTest/java/net/grandcentrix/thirtyinch/internal/TiActivityDelegateTest.java
```diff
@@ -1,276 +0,0 @@
...
-import android.support.test.runner.AndroidJUnit4;
-import android.test.suitebuilder.annotation.SmallTest;
-
-import java.util.concurrent.Executor;
-
-import static junit.framework.Assert.assertEquals;
-
-@RunWith(AndroidJUnit4.class)
-@SmallTest
-public class TiActivityDelegateTest {
```
passsy: half of the tests got removed because `setUseStaticSaviorToRetain` is now deprecated and always `true`. The other half was duplicated and already covered with JVM tests

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
passsy: currently for every `Activity` a new `HashMap` is required to store the presenters and to clean all when the `Activity` get destroyed. So, every Activity has it's own scope of presenters. I called this class `ActivityScopedPresenters`. I'm very open for better names. It was late when I wrote this

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
passsy: more `mActivityIdForPresenters`. I actually like `mScopes` here 😉

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
passsy: The `Savior` creates a new scope (`HashMap`) for every `Activity` where a Presenter is used. The `id` of that scope is saved in the `Activity`. When the Activity recreates it receives its scope where the Activity can access a presenter using the `presenterId`. There is no `activityId`.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+    private synchronized ActivityScopedPresenters getScope(final Activity activity) {
+        final String id = mScopeIdForActivity.get(activity);
+        if (id != null) {
+            return mScopes.get(id);
+        }
+
+        return null;
+    }
+
+    private synchronized ActivityScopedPresenters getScopeOrCreate(final Activity activity) {
```
StefMa: Please use `getScope()` here as well and/or merge these two methods anyhow... They do the same thing

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+    public void saveScopeId(final Activity activity, final Bundle outState) {
+        String id = mScopeIdForActivity.get(activity);
+        if (id == null) {
+            id = UUID.randomUUID().toString();
+            mScopeIdForActivity.put(activity, id);
+        }
+        outState.putString(TI_ACTIVITY_PRESENTER_SCOPE_KEY, id);
+    }
+
+    private synchronized ActivityScopedPresenters getScope(final Activity activity) {
```
StefMa: Add `@Nullable`

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
StefMa: Add `@Nonull`

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+            }
+
+            @Override
+            public void onActivityStopped(final Activity activity) {
+
+            }
+        };
+    }
+
+    public void saveScopeId(final Activity activity, final Bundle outState) {
```
StefMa: Should not be public!

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -32,36 +39,261 @@
...
+        }
+        return INSTANCE;
+    }
+
+    @VisibleForTesting
+    /*package*/ PresenterSavior() {
+
+    }
+
+    public void detectNewActivity(final Activity activity, final Bundle savedInstanceState) {
```
StefMa: Should not be public!

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
StefMa: Thank you for that explanation.
No i got it :)

The question is:
Why do we need a extra class for it?
Either it can be directly a HashMap (`extends HashMap`) or we can remove it completelly... 🤔

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
passsy: `@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)` private is already the default

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
passsy: sure we could remove this class but I thought working with a 
`HashMap<String, HashMap<String, TiPresenter>>` is not the best idea

gborland:
I'm the original reporter of #33 and I'm happy to confirm that this fixes the issue for me. 👍

passsy:
I was able to simplify the `onDestroy()` logic in `TiFragment` and `TiActivity`. I also added all required tests for the `TiActivity`.

Also:
- removed the `NonConfigurationInstance` recover method. `setUseStaticSaviorToRetain(false)` is deprecated and therefore it was never used. The savior does a better job.
- removed hard "Don't keep Activities" checks. This is not required anymore. `isFinishing` and `isChangingConfigurations` are enough information. The "Don't keep Activities" check was unreachable code.
- `TiFragment#setRetainInstanceState(true)` will now throw and is not supported. 
- added documentation

I'm now happy with the result and look forward for the merge. Next step: write release notes 😁

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
ar-g: ActivityScopedPresenters not necessary but it makes code more readable

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Simple wrapper around a {@link HashMap} to save {@link TiPresenter} by id. For every {@link
+ * android.app.Activity} containing a {@link TiPresenter} a corresponding {@link
+ * ActivityScopedPresenters} will be created
```
StefMa: Please add the info, that this will store also the Presenters for the `Fragments` which are added to `that Activity`.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/ActivityScopedPresenters.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class ActivityScopedPresenters {
```
StefMa: I think about implementing `Map` directly or extending from `HashMap`. But not needed now. So keep it like it is ;)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -15,53 +15,226 @@
...
+     * destroyed. The {@link ActivityInstanceObserver} takes care to manage unique Ids for each
+     * Activity which are used as keys here.
+     */
+    @VisibleForTesting
+    final HashMap<String, ActivityScopedPresenters> mScopes = new HashMap<>();
+
+    /**
+     * Access to the {@link PresenterSavior} singleton to save presenters across orientation changes
+     *
+     * @return singleton for
```
StefMa: remove that. I guess that is clear :D

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -15,53 +15,226 @@
...
+        if (mScopes.isEmpty()) {
+            // unregister detector because there are no presenters which could be recovered.
+            // next #save call will create a new one
+            if (mActivityInstanceObserver != null) {
+                if (DEBUG) {
+                    TiLog.v(TAG, "unregistering lifecycle callback");
+                }
+                activity.getApplication()
+                        .unregisterActivityLifecycleCallbacks(mActivityInstanceObserver);
+                mActivityInstanceObserver = null;
```
StefMa: Dublicated code (see #free).
Please extract it in `destroyActivtiyInstanceObserver` or something

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -15,53 +15,226 @@
...
+        }
+
+        printRemainingPresenter();
+    }
 
     @Override
-    public void free(final String presenterId) {
-        mPresenters.remove(presenterId);
+    public void onActivityFinished(final Activity activity, final String activityId) {
+        final ActivityScopedPresenters scope = mScopes.remove(activityId);
```
StefMa: Move that below the statement `mScope.isEmpty()`. We need it only after this statement...

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
```diff
@@ -15,53 +15,226 @@
...
+     *
+     * @return an existing or new created scope for presenters
+     */
+    @NonNull
+    private synchronized ActivityScopedPresenters getScopeOrCreate(final Activity activity) {
+        if (mActivityInstanceObserver == null) {
+            mActivityInstanceObserver = new ActivityInstanceObserver(this);
+            TiLog.v(TAG, "registering lifecycle callback");
+            activity.getApplication().registerActivityLifecycleCallbacks(mActivityInstanceObserver);
+        }
```
StefMa: Please extract this statement to a method like `registerObserverAsLifecyclerCallback`

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterSavior.java
passsy: the scope is only required in on branch, true. But removing the scope is the important part here. required for both branches.

StefMa: Then maybe moving up the second statement? 🤔

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiFragment.java
```diff
@@ -24,21 +24,90 @@
...
+ * The {@link TiPresenter} requires a interface to communicate with the View. Normally the Activity
+ * implements the View interface (which must extend {@link TiView}) and is returned by default
+ * from {@link #provideView()}.
+ * </p>
+ * <p>
+ * The associated {@link TiPresenter} only lives when the {@link TiFragment} is added to the
+ * {@link FragmentManager}. When {@link FragmentTransaction#remove(Fragment)} or {@link
+ * FragmentTransaction#replace(int, Fragment)} results in removing this {@link TiFragment} from the
+ * {@link FragmentManager} the {@link TiPresenter} gets destroyed. When the same {@link TiFragment}
+ * instance will be added again a new {@link TiPresenter} will be created by calling {@link
```
passsy: I added this case to the fragment lifecycle test Activity and can't reproduce it. The presenter is always destroyed and a new one is generated.

passsy:
multiple calls to `TiPresenter#destroy()` are ok. We have multiple triggers to detect the Activity destroy case and we can't say which one fires first. We could guard them with `if (!presenter.isDestroyed()) { presenter.destroy(); }`. I don't think this is useful.

StefMa:
I've checked the ViewPager behaviour like we discussed.
See #84

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiFragment.java
StefMa: That was discussed and had the right behaviour.
The second (or following) added fragment got the same presenter.
Why? Because we added it to the backstack. And the default behaviour of `TiFragment`s-Presenter is to keep as long as you are managed by the fragment manager. And if you are in the backstack, you are managed by them ...

Issue #29, opened 2016-10-29
@DistinctUntilChanged doesn't clear cache in TiFragment after orientation change


passsy:
Reproduce:
1. call `showGallery(new ArrayList())` TiFragment implementing this view interface method

``` java
    @DistinctUntilChanged
    void showGallery(List<Image> images);
```
1. make orientation change
2. call view again with same argument (`new ArrayList()`)

_Expected behavior:_
View method gets called

_Actual behavior:_
View method doesn't get called because it's the same data. This is wrong because `onCreateView()` was called and a new view implementation was created.

Workaround:

clear the cache manually in `onCreateView()`

``` java
final DistinctUntilChangedInterceptor distinctUntilChangedInterceptor = (DistinctUntilChangedInterceptor)
                getInterceptors(interceptor -> interceptor instanceof DistinctUntilChangedInterceptor).get(0);
        distinctUntilChangedInterceptor.clearCache(this);
```

rberghegger:
I tried to reproduce this issue and noticed that the bug does seem to be fixed.
Please notice that it is very important to call `super.onCreateView()` in your `onCreateView()` implementation because otherwise `TiFragment` is not able to invalidate the cache of the latest bound view itself.

I would suggest to close this issue and look into my pull request which restricts the `TiFragment` api to subclasses and enforces the call of the super implementation when overriding important lifecycle methods.

StefMa:
I can reproduce the error on "not calling super" as @rberghegger mentioned.

So I agree with @rberghegger.
Closed and see #79

PR #79, opened 2017-04-04
Feature: restrict TiFragment api for subclasses

rberghegger:
This pull request restricts the `TiFragment` api by declaring some implemented methods from the `DelegatedTiFragment`-Interface as `final` so they cannot be overridden by subclasses.
It also adds the `@CallSuper`-Annotation to all important lifecycle methods that requires calling the super implementation because `TiFragment` uses this lifecycle methods to manage the `TiFragmentDelegate` instance.

This should help to prevent issues which are caused by extending `TiFragment` and not call its respective super implementations (see #29 for example).

@passsy suggested to use the `FragmentManager.FragmentLifecycleCallbacks` (https://developer.android.com/reference/android/support/v4/app/FragmentManager.FragmentLifecycleCallbacks.html) to enforce the calls on the managed `TiFragmentDelegate` instance but since we are trying to hit the 0.8.0 release i think this would be a rather drastic improvement.

### DIFF
--- a/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
+++ b/plugin/src/main/java/net/grandcentrix/thirtyinch/plugin/TiActivityPlugin.java
@@ -38,6 +38,7 @@ import net.grandcentrix.thirtyinch.util.AnnotationUtil;
 import android.app.Activity;
 import android.content.res.Configuration;
 import android.os.Bundle;
+import android.support.annotation.CallSuper;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


passsy:
Do the same for `TiActivity` as well as `TiFragmentPlugin` and `TiActivityPlugin`

rberghegger:
@passsy I'm not sure whether or not `onRetainCustomNonConfigurationInstance` in the `TiActivity`-Class should be finalized or annotated with `@CallSuper` as well. 🤔

passsy:
`final` would be wrong. Add `@CallSuper`

rberghegger:
I applied your suggestions to the `TiActivity`, `TiActivityPlugin` and the `TiFragmentPlugin`.

passsy:
`TiDialogFragment` is missing 😉

PR #80, opened 2017-04-06
Feature: Add a LoggingInterceptor

rberghegger:
Adds a `LoggingInterceptor` which implements the `BindViewInterceptor`-Interface and can be used to monitor the method calls on a bound view.

This Class was originally implemented by @passsy and I made only a few changes:
instead of accessing the build config to determine if the logging should be enabled this is now controlled via the flag `mEnableLogging` which is set via the classes constructor.


passsy:
Should we put it in a separate module like [okhttp-logging-interceptor](https://github.com/square/okhttp/tree/master/okhttp-logging-interceptor)? Does anyone has a option on this?

StefMa:
Yes, I've already thought about it.
It don't want such a Interceptor in the ThirtyInch codebase directly...
A separate module would be great... But.. I don't know if that is to much overhead for that single class. I don't know if they worth it :)

passsy:
I added an updated version with improved logging I found in a stale branch. And added some tests :)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiLog.java
```diff
@@ -99,6 +125,10 @@ public static void i(final String tag, final String msg) {
      * </code>
      */
     public static void setLogger(@Nullable final Logger logger) {
+        if (logger == TI_LOG) {
+            throw new IllegalArgumentException(
+                    "Recursion warning: You can't use TI_LOG as Logger for TiLog");
```
StefMa: If we can't do it. Why do we have it then? 🤔 
Can be removed? 🌮

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiLog.java
```diff
@@ -54,8 +54,28 @@ public void log(final int level, final String tag, final String msg) {
...
+
     private static Logger logger;
 
+    /**
+     * forward log to {@link TiLog} for logging
+     */
+    public static Logger TI_LOG = new Logger() {
+        @Override
+        public void log(final int level, final String tag, final String msg) {
+            TiLog.log(level, tag, msg);
```
StefMa: I recomment to remove the "self login".
Instead i would add a `System.out.print(level, tag, msg)` (or smth)

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiLog.java
passsy: It's a implementation of this interface which can be used everywhere except for this location which would result in recursion.
I really like that the `LoggingInterceptor` logs to `TiLog` by default using the `TiLog.TI_LOG`.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/LoggingInterceptor.java
```diff
@@ -0,0 +1,189 @@
...
+                    .toString();
+        }
+
+        @Override
+        protected Object handleInvocation(final Object proxy, final Method method,
+                final Object[] args)
+                throws Throwable {
+
+            try {
+                mLogger.log(Log.VERBOSE, TAG, toString(method, args));
```
passsy: yes, the `InvocationHandler` is just a implementation detail. You add a `LoggingInterceptor` and you expect this interceptor to log something.

StefMa:
Because it have some conflicts I've rebased it in #85 and changed a variable.
See #85 for now

PR #81, opened 2017-04-13
Minimize correctly with Proguard

passsy:
## Description

Bundles the proguard rules required for `Ti` within the `aar` artifacts. ThirtyInch now support Proguard without additional adjustments 

## How to test
- Enable proguard in app (sample app here is fine) and build. Everything works
- remove `rxjava:proguard-rules` dependency, build will fail
- remove keep rule for `DistinctComparator` will crash at runtime (app startup)

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -16,6 +16,13 @@ android {
         versionName "1.0"
     }
     buildTypes {
+        debug {
+            // for testing proguard
+            //
+            // minifyEnabled true
+            // useProguard true
+            // proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
         release {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -17,15 +17,18 @@ android {
     }
     buildTypes {
         debug {
+            // only use the build in shrinker for debug builds, don't obfuscate
             // remove unused classes and to make sure the proguard config is correct
+            // http://tools.android.com/tech-docs/new-build-system/built-in-shrinker
             // https://developer.android.com/studio/build/shrink-code.html#shrink-code
             minifyEnabled true
-            // don't enable proguard, it doesn't work with instant run
+            // using the build in shrinker only works with proguard disabled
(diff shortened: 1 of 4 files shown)


sample/build.gradle
```diff
@@ -16,6 +16,13 @@ android {
         versionName "1.0"
     }
     buildTypes {
+        debug {
+            // for testing proguard
```
StefMa: Why not enabling it? ;)

passsy: It increases the compile time. But you're right, this sample is so small it doesn't matter

sample/build.gradle
```diff
@@ -16,8 +16,17 @@ android {
...
+            // https://developer.android.com/studio/build/shrink-code.html#shrink-code
+            minifyEnabled true
+            // don't enable proguard, it doesn't work with instant run
+            // https://developer.android.com/studio/build/shrink-code.html#gradle-shrinker
+            useProguard false
+            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+        }
         release {
-            minifyEnabled false
+            useProguard true
```
StefMa: We have to use `minifyEnabled true`. Not `useProguard true` as I understand the documentation right...

sample/build.gradle
```diff
@@ -16,8 +16,17 @@ android {
         versionName "1.0"
     }
     buildTypes {
+        debug {
+            // remove unused classes and to make sure the proguard config is correct
+            // https://developer.android.com/studio/build/shrink-code.html#shrink-code
+            minifyEnabled true
+            // don't enable proguard, it doesn't work with instant run
+            // https://developer.android.com/studio/build/shrink-code.html#gradle-shrinker
+            useProguard false
```
StefMa: This is very strange. Because this config don't create the output of the mapping files etc.
But it is implemented like documented .. so 🤷‍♂️

sample/build.gradle
StefMa: Is seems that useProguard is not really a "public" api...
http://google.github.io/android-gradle-dsl/2.3/com.android.build.gradle.internal.dsl.BuildType.html

sample/build.gradle
passsy: it only removed unused code and doesn't obfuscate the code

PR #86, opened 2017-04-26
Use hosting activity in TiFragment

vRallev:
This gives subclasses a chance to override the activity.

### DIFF
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiFragment.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiFragment.java
@@ -189,12 +189,12 @@ public abstract class TiFragment<P extends TiPresenter<V>, V extends TiView> ext
 
     @Override
     public final boolean isHostingActivityChangingConfigurations() {
-        return getActivity().isChangingConfigurations();
+        return getHostingActivity().isChangingConfigurations();
     }
 
     @Override
     public final boolean isHostingActivityFinishing() {
-        return getActivity().isFinishing();
+        return getHostingActivity().isFinishing();
     }
 
     @Override


passsy:
Can you give me a specific usecase? For tests you can use [`TestTiFragment`](https://github.com/grandcentrix/ThirtyInch/blob/master/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/TestTiFragment.java)

vRallev:
Let's say I want to inject a mock activity into a fragment. I thought the idea would be that the hosting activity should be used in all cases. Otherwise I don't see a reason why this method even exists.

passsy:
`TiFragment` is not designed to be tested, that's why we wrote the `TiFragmentDelegate` which contains all the logic. This delegate can be tested with `TestTiFragment`. You may have to add a setter for the HostingActivity for the `TestTiFragment.Builder`.

`getHostingActivity()` is primary designed to return the Activity instance (i.e. `mock(Activity.class)`) for equals comparisons inside of the `PresenterSavior`. It wasn't added to call methods on it.

vRallev:
Maybe I'm missing something, but then the method names are confusing and not self explaining. I'm exactly having this case where I override `get**HostingActivity**()` and return a mock. But it's weird to me, that `is**HostingActivity**Finishing()` suddenly uses another activity and not what I return.

passsy:
Yes it's a naming and documentation problem. Also `getHostingActivity()` should be `final` preventing you from running in this problem and forcing you to use `TestTiFragment` in the first place. 😏

vRallev:
In this case you can remove the method and call `getActivity()` directly.

passsy:
The `TiFragmentDelegate` doesn't know the `Fragment` to call `getActivity()`. It only knows `DelegatedTiFragment` which has the `getHostingActivity()` method.

passsy:
fixed with #88

PR #89, opened 2017-04-28
fixed manageSubscriptions where it should call manageSubscription

k0shk0sh:
instead of manageViewSubscription in RxTiPresenterSubscriptionHandler


StefMa:
Oh wow!
Nice catch. Thank you 👍 

Can you please add a test so that something like that will never happen again?!
You can simple copy tests like [this](https://github.com/grandcentrix/ThirtyInch/blob/master/rx/src/test/java/net/grandcentrix/thirtyinch/rx/RxTiPresenterSubscriptionHandlerTest.java#L192) but called `manageSubscriptions`.

k0shk0sh:
@StefMa sure, I'll write up that right now.

k0shk0sh:
@StefMa didn't want to introduce something new, so I followed up `manageSusbription` tests.

passsy:
Ouch, that was a good catch!

passsy:
@StefMa please add a test for [RxTiPresenterDisposableHandler#manageDisposables(Disposable...)](https://github.com/grandcentrix/ThirtyInch/blob/master/rx2/src/main/java/net/grandcentrix/thirtyinch/rx2/RxTiPresenterDisposableHandler.java#L84)

PR #90, opened 2017-04-28
Added tests for manageDisposables

StefMa:
According to @passsy in #89 **I have to** add some tests for `manageDisposables`.

et voilà


passsy:
Thanks 👍

PR #93, opened 2017-05-03
Ti logo

passsy:
First draft of a logo for `Ti`. Uses `Roboto Mono` as font


vRallev:
Could you create a version without the `Ti` and use the font for the `30`?

passsy:
`Ti` and `30` is both `Roboto Mono`

passsy:
This one could be quite confusing 😛 

I'm already thinking about something else 🤔  💭

StefMa:
I like it. I like booth (maybe because the first one was my idea? 😉)

PR #94, opened 2017-05-03
Always call observer for events which happend when they were registered

passsy:
It happend that an observer removes another observer. This resulted in not calling the removed observer when the observer was registered after the removing observer.

Basically, don't change the list while iterating over it.

### DIFF
--- a/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiLifecycleObserverTest.java
+++ b/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiLifecycleObserverTest.java
@@ -33,8 +33,7 @@ import static org.mockito.Matchers.anyBoolean;
 import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.times;
-import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoMoreInteractions;
 
 public class TiLifecycleObserverTest {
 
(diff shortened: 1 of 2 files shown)


thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiLifecycleObserverTest.java
```diff
@@ -249,6 +256,33 @@ public void onChange(final TiPresenter.State state,
...
+            public Object answer(final InvocationOnMock invocation) throws Throwable {
+                removable.remove();
+                return null;
+            }
+        }).when(observer1).onChange(any(TiPresenter.State.class), anyBoolean());
+
+        mPresenter.attachView(mock(TiView.class));
+
+        //receives both events
+        verify(observer1, times(2)).onChange(any(TiPresenter.State.class), anyBoolean());
```
StefMa: For understadning:
We call `attachView` which triggers the `fori`.
The list have two items. Observable1 and Observable2.
`observable1#onChange` got called with `init` and `attach`. Which triggers the answer to remove the `observable2`from the Observable list.
But because of the fact we have a local copy, `observable2#onChange` got also called with `init`. But not with `attach` anymore...

Right? 🤔

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiLifecycleObserverTest.java
```diff
@@ -249,6 +256,33 @@ public void onChange(final TiPresenter.State state,
...
+
+        // only gets first event, already unregistered when second event fires
```
StefMa: Maybe we have to adjust the comment then. `Only gets the init event. Aldready unregistered for attach event`...

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiLifecycleObserverTest.java
StefMa: Are ok.. 👍 
But I would recommend to change the comment (https://github.com/grandcentrix/ThirtyInch/pull/94#discussion_r114708118)  with that info.. Instead of just "first event" and "second event"...

Maybe you can also improve the tests with the given State and Boolean value 🤔

Issue #95, opened 2017-05-04
Fix JavaDoc issues


StefMa:
We have noticed in #88 that we have some javadoc issues.
See [this comment](https://github.com/grandcentrix/ThirtyInch/pull/88#issuecomment-298845275). (Maybe there are more than just this 😉)

We have to fix that.

StefMa:
Fixed by #99

_Sent from my LGE Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

PR #88, opened 2017-04-26
Feature/presenter host refactoring

passsy:
More or less a cleanup PR for #78 

@vRallev pointed out #86 that the naming in `DelegatedTiFragment` is quite confusing. 
- `getHostingActivity()` returns `Activity` but shouldn't used as `Activity` to call methods like `isFinishing()`. 
- `isHostingActivityFinishing()` doesn't read the `isFinishing()` property from `getHostingActivity()`
- `isHostingActivityChangingConfigurations()` doesn't read the `isChangingConfigurations()` property from `getHostingActivity()`

Solution:
- `isHostingActivityFinishing()` and `isHostingActivityChangingConfigurations()` can be removed. The `PresenterSavior` is now 100% responsible to destroy the presenter when a `Activity` containing a `TiFragment` finishes.
- `Activity getHostingActivity()` was replaced with `Object getHostingContainer()` in `DelegatedTiFragment` (returning `Fragment#getHost()`) and `DelegatedTiActivity` (returning itself)

Also:
- `PresenterSavior` calls now all `TiPresenter` destructing methods including `detachView()`
- Fixed bug where `TiPresenter#destroy()` didn't destroy the presenter when the view is attached
- Removed `TestTiFragment.Builder#setDontKeepActivitiesEnabled(boolean)` which actually did nothing.
- Fix missing overrides for `TiDialogFragment`

PR #99, opened 2017-05-30
Fix some javadoc warnings

StefMa:
This will fixes some javadoc warnings we found in #88 
See also #95 

There is one issue left which says:
> src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java:515: warning - Tag @link: reference not found: android.widget.Toast

But I'm unable to fix that 🤷‍♂️
If you can give me a hint - feel free 👍


passsy:
The `Toast` `import` is missing in `TiPresenter`

PR #105, opened 2017-07-10
Deprecate/remove changing configuration

StefMa:
This will declare `DelegatedTiActivity#isActivityChangingConfigurations` as deprecated. We don't need that call anymore. Our internal API have removed the call to it. 
Public API's are set to deprecated with a hint what to use now.

With that change I also removed the generic from the `DelegatedTiActivtiy`. It's not needed...


passsy:
Don't deprecate it, remove it completely. It's part of the `internal` package, we can change it at any time

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/MultipleTiFragmentPresenterDestroyTestIgnoreDontKeepActivities.java
```diff
@@ -139,9 +138,6 @@ public void activityChangingConfiguration_retainTrue_backstackFalse_dkATrue() {
         assertThat(fragment.getPresenter().isDestroyed()).isTrue();
         assertThat(mSavior.getPresenterCount()).isEqualTo(0);
 
-        // When the Activity is changing its configuration.
```
passsy: this comment should stay although the activity parameters don't change

PR #106, opened 2017-07-16
deprecated onWakeUp() removed

Rainer-Lang:
I'm not sure about possible side effects....


StefMa:
Thank you for the contribution @Rainer-Lang.

`onWakeUp()` was defined in first iteration of `Ti`.
We decided that the name don't really match well. So we have replaced it with `onAttachView(view)` in the [latest release](https://github.com/grandcentrix/ThirtyInch/releases/tag/v0.8.0).

To don't break other apps which uses `ThirtyInch` bevor 0.8.0 but update to it, we decided to leave it in but in a `Deprecated` state.

With that knowledge I hope you understand why we can't merge your PR...

Rainer-Lang:
Ok. I only saw the log and thought "why I'm getting this, I don't use it....." 😉

PR #108, opened 2017-07-19
Let's detect Ti by AboutLibraries

StefMa:
Since some apps (e.g. [FastHub](https://github.com/k0shk0sh/FastHub)) uses ThirtyInch along with [AboutLibraries](https://github.com/mikepenz/AboutLibraries) we should add the support for it.

This PR will fix it. If someone uses ThirtyInch as a library with AboutLibraries it will be automatically added to the AboutLibrary-Activity.

You can test it by checking our [this branch](https://github.com/StefMa/ThirtyInch/tree/test/about_libraries).

A gif in action:
geist_1500457576288

### DIFF
--- a/thirtyinch/src/main/res/values/library_thirtyinch_strings.xml
+++ b/thirtyinch/src/main/res/values/library_thirtyinch_strings.xml
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="define_thirtyinch">year;owner</string>
+    <string name="library_thirtyinch_author">grandcentrix</string>
+    <string name="library_thirtyinch_authorWebsite">https://www.grandcentrix.net/</string>
+    <string name="library_thirtyinch_classPath">net.grandcentrix.thirtyinch.TiPresenter</string>
+    <string name="library_thirtyinch_libraryName">ThirtyInch</string>
+    <string name="library_thirtyinch_libraryDescription">A MVP library for Android favoring a stateful Presenter.</string>
+    <string name="library_thirtyinch_libraryVersion">0.8.0</string>
+    <string name="library_thirtyinch_libraryWebsite">https://github.com/grandcentrix/ThirtyInch</string>
+    <string name="library_thirtyinch_licenseId">apache_2_0</string>
+    <string name="library_thirtyinch_isOpenSource">true</string>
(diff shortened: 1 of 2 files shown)


thirtyinch/src/main/res/values/library_thirtyinch_strings.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="define_thirtyinch">year;owner</string>
+    <string name="library_thirtyinch_author">grandcentrix</string>
+    <string name="library_thirtyinch_authorWebsite">https://www.grandcentrix.net/</string>
+    <string name="library_thirtyinch_classPath">net.grandcentrix.thirtyinch.TiPresenter</string>
+    <string name="library_thirtyinch_libraryName">ThirtyInch</string>
+    <string name="library_thirtyinch_libraryDescription">A MVP library for Android favoring a stateful Presenter.</string>
+    <string name="library_thirtyinch_libraryVersion">0.8.0</string>
```
passsy: can we use some kind of placeholders?

passsy:
I think `AboutLibraries` uses the wrong technology. It should be a gradle plugin parsing the pom files of each artifact for the required information. 🤔

passsy:
Any why would someone include

passsy:
Ups, missclick

StefMa:
I'm totally with you.
I don't like the implementation of AboutLibraries as well. 

I just noted that FastHub don't included ThirtyInch in the Open Source licenses. But I wanted that the include us 😄 
That was the only reason why I created that PR :D

> can we use some kind of placeholders?

Yep, we can introduce some placeholders here. We can create the strings inside the `build.gradle` for example.

> And also when somebody includes AboutLibraries tons of unused string resource will be included. Just checked the proguarded APK

Yeah true. I've tried to just include the `define_thirtyinch` plus `librarie_thirtyinch_*` but without success. If you found a solution with proGuard which remove all unsused strings but not the strings above... Would be great :D

passsy:
Create the PR to the FastHub project 😉 

We already provide all information with our `pom.xml`. There are [libraries](https://github.com/cookpad/license-tools-plugin) using it. `AboutLibraries` should use the standard way to get those information at compile time instead of cluttering all libraries and apps with unused resources.
// cc @mikepenz

Issue #100, opened 2017-06-15
ThirtyInch Lint Checks


mannodermaus:
ThirtyInch has quickly become my favourite MVP implementation. Still, and I guess it's mostly due to my forgetful mind, nearly *every time* I implement a new Ti-based `Activity`, connect everything & run, I encounter the dreaded error I've grown so accustomed to:
`java.lang.IllegalArgumentException: This Activity doesn't implement a TiView interface. This is the default behaviour. Override provideView() to explicitly change this.`

I'd like to propose some custom Lint checks that facilitate the daily workflow with ThirtyInch. I'm unsure how feasible this would be, especially since you allow this default behaviour to be changed on a per-Activity basis. The error would ideally be triggered on `TiActivity` & `CompositeActivity` sub-classes with the `TiActivityPlugin` applied, if no `TiView` interface is implemented.

Again, I'm probably the only one oblivious enough to always forget this step, and maybe the effort outweighs the benefit by a long shot.

passsy:
Yeah that would be a great addition. This lint check could check if `provideView()` is overriden, if not it should check if the Activity/Fragment implements the `View` interface.

I must admit I have no idea how to write custom lint rule and bundle it within the `aar`.

passsy:
For inspirations see [timber lint checks](https://github.com/JakeWharton/timber/tree/master/timber-lint)

//cc @winterDroid @vanniktech

winterDroid:
Can you provide some more information what issues should be found? I'm not really familiar with this library

vanniktech:
Why was I cc'd into this?

passsy:
@vanniktech You're familiar with lint and maybe interested in contributing 😉, ignore otherwise

@winterDroid This is the minimal setup when using a `TiActivity`

```java
// Lint warning because MyActivity doesn't implement MyView 
public class MyActivity extends TiActivity<MyPresenter, MyView> {

    @Override
    public MyPresenter createPresenter() {
        return new MyPresenter();
    }
}

public class MyPresenter extends TiPresenter<MyView> {
    // boilerplate
}

interface MyView extends TiView {
    // boilerplate
}
```

It compiles fine but [crashes](https://github.com/grandcentrix/ThirtyInch/blob/master/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiActivity.java#L179) at runtime because `MyActivity` doesn't implement `MyView`. Lint should warn here.

Correct `MyActivity` implementation:

```java
[...]

PR #111, opened 2017-07-23
Added Lint Check for missing TiView implementation

mannodermaus:
This is an implementation based on the proposal made in #100. It introduces custom lint checks for ThirtyInch with a [UAST-based](https://groups.google.com/forum/#!topic/lint-dev/7nLiXa04baM) `Detector` that seeks to point out illegal configuration of `TiActivity`, `TiFragment`, `CompositeActivity` & `CompositeFragment` sub-classes. Basically, it works like this:

* Visit every non-abstract sub-class of either of the classes mentioned above
* Retrieve the `TiView` sub-interface that the sub-class is connected to
* Validate that the sub-class implements this `TiView` by checking its signature

Note that there is an exception to this rule for `Ti-` classes, wherein a user can override `TiViewProvider#provideView()` to deviate from the default behaviour. The Detector verifies that, and doesn't report an issue on sub-classes that also override this method.

Integration with CompositeAndroid works in a more low-level fashion, where the default constructor of the `Composite-` sub-class is analyzed for its registration of a `TiPlugin` via `addPlugin()`. If found, the `TiView` is extracted from there.

Please take a close look at the behaviour of the Detector, and also at potential loopholes inside the unit tests.


lint/src/main/java/net/grandcentrix/thirtyinch/MissingTiViewImplementationDetector.java
```diff
@@ -0,0 +1,250 @@
...
+            "net.grandcentrix.thirtyinch.TiActivity",
+            "net.grandcentrix.thirtyinch.TiFragment"));
+    private static final String ADD_PLUGIN_METHOD = "addPlugin";
+    private static final String TI_ACTIVITY_PLUGIN_NAME = "TiActivityPlugin";
+    private static final String TI_FRAGMENT_PLUGIN_NAME = "TiFragmentPlugin";
+    private static final List<String> CA_CLASS_NAMES = unmodifiableList(asList(
+            "com.pascalwelsch.compositeandroid.activity.CompositeActivity",
+            "com.pascalwelsch.compositeandroid.fragment.CompositeFragment"));
+
+    static final Issue ISSUE = Issue.create(
```
aurae: This should probably be split into two separate `Issue` descriptions with the same ID, which provide distinct descriptions for `Ti` and `Composite` use cases. Otherwise, the part about `provideView()` is just confusing.

build.gradle
```diff
@@ -34,4 +34,6 @@ ext {
     junitVersion = '4.12'
     mockitoVersion = '1.10.19'
     assertjVersion = '2.5.0'
+
+    lintVersion = '25.4.0-alpha7'
```
passsy: Is the alpha version required for UAST? Isn't there a stable one?

lint/src/main/java/net/grandcentrix/thirtyinch/MissingTiViewImplementationDetector.java
passsy: Yep normal Ti and composite should be separated, especially because we may get rid of composite #109

lint/build.gradle
```diff
@@ -0,0 +1,30 @@
...
+    maven { url "https://dl.bintray.com/android/android-tools" }
+}
+
+dependencies {
+    compile "com.android.tools.lint:lint-api:$lintVersion"
+    compile "com.android.tools.lint:lint-checks:$lintVersion"
+
+    testCompile "com.android.tools.lint:lint:$lintVersion"
+    testCompile "com.android.tools.lint:lint-tests:$lintVersion"
+    testCompile 'org.easytesting:fest-assert-core:2.0M10'
```
passsy: please use AssertJ as in all other modules

build.gradle
aurae: The latest stable release of the Lint libraries (25.3.0) doesn't include UAST yet, no. I suppose it'll be out once AS 3.0 is stable, though.

aurae:
Thanks for the initial feedback! I'd love to rework the module into Kotlin, as it would absolutely allow for cleaner structure in several places.

aurae:
I have separated the Detector into two distinct cases for Ti and CompositeAndroid. The implementations themselves are in Kotlin now, as well. The only remaining piece of Java code is present in the unit tests for the Detectors, pretty much only because of IJ's nice syntax highlighting on `@Language("JAVA")` strings. Ready for review!

lint/src/main/kotlin/net/grandcentrix/thirtyinch/Issues.kt
```diff
@@ -0,0 +1,31 @@
+package net.grandcentrix.thirtyinch
+
+import com.android.tools.lint.detector.api.*
+import java.util.*
+
+enum class Issues(
```
StefMa: I don't like enum. Especially if we can just create a`data class` instead.
Addional, I don't see the reason why it is a enum. We cal also creta a "normal" class,,

lint/src/test/java/net/grandcentrix/thirtyinch/MissingViewInThirtyInchDetectorTest.java
```diff
@@ -0,0 +1,261 @@
...
+            "package net.grandcentrix.thirtyinch;\n" +
+            "public abstract class TiPresenter<V extends TiView> {\n" +
+            "}");
+
+    private final TestFile tiViewStub = java("" +
+            "package net.grandcentrix.thirtyinch;\n" +
+            "public interface TiView {\n" +
+            "}");
+
+//    private final TestFile caBasePluginStub = java("" +
```
StefMa: Why do we have outcommented tests?
Can be removed then, right?

lint/src/main/kotlin/net/grandcentrix/thirtyinch/MissingViewInCompositeDetector.kt
```diff
@@ -0,0 +1,94 @@
+package net.grandcentrix.thirtyinch
+
+import com.android.annotations.VisibleForTesting
+import com.android.tools.lint.detector.api.Issue
+import com.android.tools.lint.detector.api.JavaContext
+import com.intellij.psi.PsiClass
+import com.intellij.psi.PsiClassType
+import com.intellij.psi.PsiJavaCodeReferenceElement
+import com.intellij.psi.PsiType
+import org.jetbrains.uast.*
```
StefMa: Please don't import "everthing" from the `uast` package. Just the things we need.

lint/src/main/kotlin/net/grandcentrix/thirtyinch/Issues.kt
aurae: The `Issues` type is supposed to declare the limited set of available detections that the ThirtyInch Lint module can analyze. Both `data` and "normal" classes wouldn't easily allow for this defined set of options, unless we used constants, which I wouldn't encourage. A `sealed class` could work, yet each implementation would have the same duplicated properties anyway.

lint/src/test/java/net/grandcentrix/thirtyinch/MissingViewInThirtyInchDetectorTest.java
aurae: In the process of separating Lint detection around the different available use cases, these were left behind.

lint/src/main/kotlin/net/grandcentrix/thirtyinch/MissingViewInCompositeDetector.kt
aurae: AS' default Kotlin code style encourages wildcard imports when at least 5 members of a class are being used. I have adjusted the IDE's settings and rearranged the import statements.

lint/build.gradle
```diff
@@ -0,0 +1,40 @@
...
+}
+
+jar {
+    manifest {
+        attributes('Lint-Registry': 'net.grandcentrix.thirtyinch.IssueRegistry')
+    }
+}
+
+repositories {
+    maven { url "https://dl.bintray.com/android/android-tools" }
```
StefMa: I don't like such "special cases".
But just checked. `lint-tests` isn't published to jcenter 😞 .
However. Would be great if we add this information to it.. So we can hopefully remove it if its published.

aurae: Yeah, I agree that it's unfortunate that we have to include the additional repository, just so we can access Lint-specific dependencies. A comment and, potentially, a link to an issue over at b.android.com would be helpful here!

StefMa:
I am not on my work macbook anymore. So I don't know it exactly. But I use the latest stable of android studio 2.3.3 I think 🤔?! 
And I use the same gradle plugin as you do since it is defined in the build.gradle ☺️

_Sent from my Google Nexus 5X using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

StefMa:
@aurae 
I've tried to it again with AS 3.0-Beta5.
I don't get the `NoClassDefFoundError` anymore.

But unfortunately I don't see the lint warning.
Which I've already mentioned here https://github.com/grandcentrix/ThirtyInch/pull/111#pullrequestreview-54343610

I've also tested it with AS 3.0-Beta5 with the new plugin (3.0-Beta5 of course :)):
And it doesn't work as well...

passsy:
Also interesting: https://github.com/googlesamples/android-custom-lint-rules/pull/10

aurae:
Great find @passsy, let's keep an eye on that.

Unfortunately, I'm currently unable to attend to this PR myself, so I'm unsure as to why the rule wouldn't trigger in its current state. I'm sure we will be able to fix that soon, though!

mannodermaus:
Superseded by #158

PR #115, opened 2017-09-26
Update deps and stuff

StefMa:
I've just updated all of our dependencies to the current latest version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_script:
   - adb shell input keyevent 82 &
 
 script:
-  - ./gradlew clean build jacocoTestReport assembleAndroidTest connectedCheck
+  - ./gradlew build jacocoTestReport assembleAndroidTest connectedCheck
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,11 +11,11 @@ android:
     - tools
 
     # The BuildTools version used by your project
-    - build-tools-25.0.2
+    - build-tools-26.0.1
 
     # The SDK version used to compile your project
     - android-22
-    - android-25
+    - android-26
 
     # Additional components
     - extra-google-google_play_services
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_script:
   - adb shell input keyevent 82 &
 
 script:
-  - ./gradlew build jacocoTestReport assembleAndroidTest connectedCheck
+  - ./gradlew clean build jacocoTestReport assembleAndroidTest connectedCheck
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -43,7 +43,7 @@ before_script:
   - adb shell input keyevent 82 &
 
 script:
-  - ./gradlew build jacocoTestReport assembleAndroidTest connectedCheck
+  - ./gradlew clean build jacocoTestReport assembleAndroidTest connectedCheck
```
passsy: why clean? It will delete all build artifacts from the `before_script` `assemble` step

build.gradle
```diff
@@ -24,14 +25,19 @@ ext {
...
+    // Since composite is not up to date we can't use the latest one
+    compositeAndroidVersion = supportLibraryVersion
 
     bintrayVersion = '0.5.0'
     junitVersion = '4.12'
-    mockitoVersion = '1.10.19'
-    assertjVersion = '2.5.0'
+    mockitoVersion = '2.10.0'
+    assertjVersion = '2.8.0'
+    atslVersion = '1.0.1'
```
passsy: What does `atsl` stand for?

build.gradle
```diff
@@ -24,14 +25,19 @@ ext {
...
-    BUILD_TOOLS_VERSION = '25.0.2'
+    TARGET_SDK_VERSION = 26
+    COMPILE_SDK_VERSION = 26
+    BUILD_TOOLS_VERSION = '26.0.1'
 
-    supportLibraryVersion = '25.2.0'
+    supportLibraryVersion = '26.0.0'
+    // compositeAndroidVersion should be always the same version as the supportLib version
+    // Since composite is not up to date we can't use the latest one
+    compositeAndroidVersion = supportLibraryVersion
```
passsy: good 👍

build.gradle
passsy: android test support library? I suggest `supportTestVersion`

StefMa:
I found that the travis.yml also includes `- build-tools-XX.X.X`...
Should we update these as well?
Then I will add a comment to the build-tools build.gradle as well.. Or isn't that needed? 🤔

passsy:
Yes the same comment should be added to the build tools version

build.gradle
```diff
@@ -24,14 +25,20 @@ ext {
...
-    TARGET_SDK_VERSION = 25
-    COMPILE_SDK_VERSION = 25
-    BUILD_TOOLS_VERSION = '25.0.2'
+    // If you change the SDK version don't forgot to update .travis.yml
+    TARGET_SDK_VERSION = 26
+    COMPILE_SDK_VERSION = 26
+    BUILD_TOOLS_VERSION = '26.0.1'
 
-    supportLibraryVersion = '25.2.0'
+    supportLibraryVersion = '26.0.0'
```
passsy: why not `26.1.0`?

StefMa: Because of https://github.com/grandcentrix/ThirtyInch/pull/115#discussion_r141189740 🙃

passsy: Good point!

PR #116, opened 2017-09-26
Move travis to docker

StefMa:
Based on #115 

* Do two build environments: unit and instrumentation (run in parallel) 
* Moved unit test to docker (because of the andoird images you can't really start a emulator in docker)

_**Benefits of using docker:**_
* The docker image contains only the "sdkamanger". The sdkmanager install every build-tools or android images if they needed automatically. We don't need to update the travis.yml anymore \o\
* We can trust the docker. If the underlined host (travis <-> trusty, zenty) changed, everything just works as before. As long as we can run docker in that environment.

_**Benefits of the new travisscript:**_
* It parses automatically the BUILD_TOOLS_VERSION and the SDK_VERSION from the build.gradle file and put it into the the travis android components. We don't need to update the travis.yml anymore /o/

_**Do we have speed benefits?**_
* Unfortunately not really. Because travis is very slow with starting a emulator and stuff it won't decrease the build time. Maybe 10 seconds or so because travis don't run the unit tests anymore :)

What do test:
* If codecov works as aspected -> **Checked ✅ works**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   
 android:
   components:
-    - $(if [ $TEST_SUITE == "instrumentation" ]; then echo "tools,platform-tools,tools,build-tools-26.0.1,android-22,android-26,extra-google-google_play_services,extra-google-m2repository,extra-android-m2repository,addon-google_apis-google-24,sys-img-armeabi-v7a-android-22"; fi)
+    - $(if [ $TEST_SUITE == "instrumentation" ]; then ./.travisAndroidComponents.sh; fi)
 
 before_install:
   - if [ $TEST_SUITE == "unit" ]; then
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,49 +1,42 @@
-language: android
-jdk:
-    - oraclejdk8
+sudo: required
 
+language: android
+jdk: oraclejdk8
+
+services:
+  - docker
+  
+env:
+ - TEST_SUITE=unit
+ - TEST_SUITE=instrumentation
+  
 android:
   components:
-    # Uncomment the lines below if you want to
-    # use the latest revision of Android SDK Tools
-    - tools
-    - platform-tools
-    - tools
+    - $(if [ $TEST_SUITE == "instrumentation" ]; then echo "tools,platform-tools,tools,build-tools-26.0.1,android-22,android-26,extra-google-google_play_services,extra-google-m2repository,extra-android-m2repository,addon-google_apis-google-24,sys-img-armeabi-v7a-android-22"; fi)
 
-    # The BuildTools version used by your project
-    - build-tools-26.0.1
-
-    # The SDK version used to compile your project
-    - android-22
-    - android-26
-
-    # Additional components
-    - extra-google-google_play_services
-    - extra-google-m2repository
-    - extra-android-m2repository
-    - addon-google_apis-google-24
-
-    # Specify at least one system image,
-    # if you need to run emulator(s) during your tests
-    - sys-img-armeabi-v7a-android-22
-
-
+before_install:
+  - if [ $TEST_SUITE == "unit" ]; then
+    docker pull stefma-docker-hub.bintray.io/android-build-env:0.1-alpha;
+    docker run -i -v $PWD:/project -t stefma-docker-hub.bintray.io/android-build-env:0.1-alpha /bin/bash -c "./gradlew build jacocoTestReport";
+    fi
+    
 before_script:
-
-  # First assemble the project before launching the emulator
-  - ./gradlew assemble
-
-  # List available emulators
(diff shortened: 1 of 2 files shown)


StefMa:
Seems that UnitTests are working now in docker 👍 
That is great...
Moving now InstrumentationTests back to travis...

StefMa:
Read also that: https://docs.travis-ci.com/user/speeding-up-the-build/

StefMa:
\o/ 💛  :awesome: :sweet: 🍬 

Here is 🌮

StefMa:
Next step is to cache the dependencies
https://docs.travis-ci.com/user/caching/

And other 🍬 :
* Read the android components from a file instead of having it here. Maybe we can also read the build-tools and sdk version from the `build.gradle` somehow 🤔

.travis.yml
```diff
@@ -1,49 +1,42 @@
...
-    # Specify at least one system image,
-    # if you need to run emulator(s) during your tests
-    - sys-img-armeabi-v7a-android-22
-
-
+    - $(if [ $TEST_SUITE == "instrumentation" ]; then echo "tools,platform-tools,tools,build-tools-26.0.1,android-22,android-26,extra-google-google_play_services,extra-google-m2repository,extra-android-m2repository,addon-google_apis-google-24,sys-img-armeabi-v7a-android-22"; fi)
+
+before_install:
+  - if [ $TEST_SUITE == "unit" ]; then
+    docker pull stefma-docker-hub.bintray.io/android-build-env:0.1-alpha;
```
passsy: how can we update this docker image?

StefMa:
Don't cache.
Because we can't really delete the cache ...
Make no sense with that mechanism...

.travis.yml
StefMa: It is an open source project. 
you can find it here: https://github.com/StefMa/AndroidBuildEnv

.travis.yml
```diff
@@ -1,49 +1,42 @@
...
-    - extra-google-m2repository
-    - extra-android-m2repository
-    - addon-google_apis-google-24
-
-    # Specify at least one system image,
-    # if you need to run emulator(s) during your tests
-    - sys-img-armeabi-v7a-android-22
-
-
+    - $(if [ $TEST_SUITE == "instrumentation" ]; then echo "tools,platform-tools,tools,build-tools-26.0.1,android-22,android-26,extra-google-google_play_services,extra-google-m2repository,extra-android-m2repository,addon-google_apis-google-24,sys-img-armeabi-v7a-android-22"; fi)
```
passsy: can we make the echo multiline?

StefMa: Currently I create a `script.sh` which reads the `BUILD_TOOLS_VERSION` and `SDK_VERSION` and replace the given numbers.
These script return then all components..

.travis.yml
```diff
@@ -1,49 +1,42 @@
...
+
+before_install:
+  - if [ $TEST_SUITE == "unit" ]; then
+    docker pull stefma-docker-hub.bintray.io/android-build-env:0.1-alpha;
+    docker run -i -v $PWD:/project -t stefma-docker-hub.bintray.io/android-build-env:0.1-alpha /bin/bash -c "./gradlew build jacocoTestReport";
```
passsy: move this to the `script` section

StefMa:
🤦‍♂️

StefMa:
Currently build-tools got not parsed correctly..
geist_1506584908395

StefMa:
Rebased all.
Works now 👊 

Problem was that `\'` is valid on "mac `sed`"  but not on "gnu `sed`".
Replaced it with just `'`. Works on both `sed` versions...

StefMa:
> If codecov works as aspected 

Seems that codecov is reporting fine \o/
See https://codecov.io/gh/grandcentrix/ThirtyInch/compare/7533ea03c848ab39911356fe2965e51070094d7e...5c2ba57fdf35738b40d97bff013b9565fdb2625e

.travisAndroidComponents.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Read the BUILD_TOOL_VERSION and COMPILE_SDK_VERSION
+BUILD_TOOLS_VERSION=$(cat build.gradle | grep "BUILD_TOOLS_VERSION*" | sed -e "s/^.*'\(.*\)'/\1/")
+COMPILE_SDK_VERSION=$(cat build.gradle | grep "COMPILE_SDK_VERSION*" | sed -e "s/^.*= \(.*\)$/\1/")
```
passsy: can we put this in the `.travis.yml`, too?

StefMa: No. Unfortunately this isn't possible because of the limitations of the travis.yml file

PR #117, opened 2017-09-27
Add tests for PresenterScope

passsy:
Some tests and missing nullability annotations

### DIFF
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterScope.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterScope.java
@@ -81,8 +81,8 @@ public class PresenterScope {
         // saving a presenter twice with a different id is not supported
         for (final Map.Entry<String, TiPresenter> entry : mStore.entrySet()) {
             if (entry.getValue().equals(presenter)) {
-                throw new IllegalStateException("Presenter is already saved with different id "
-                        + entry.getKey() + " " + presenter);
+                throw new IllegalStateException("Presenter is already saved with different id '"
+                        + entry.getKey() + "' " + presenter);
             }
         }
 
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterScope.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterScope.java
@@ -19,6 +19,8 @@ package net.grandcentrix.thirtyinch.internal;
 import net.grandcentrix.thirtyinch.TiLog;
 import net.grandcentrix.thirtyinch.TiPresenter;
 
+import android.support.annotation.NonNull;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -42,10 +44,12 @@ public class PresenterScope {
         return mStore.get(id);
     }
 
+    @NonNull
     public List<TiPresenter> getAll() {
         return Collections.unmodifiableList(new ArrayList<>(mStore.values()));
     }
 
+    @NonNull
     public List<Map.Entry<String, TiPresenter>> getAllMappings() {
         return Collections.unmodifiableList(new ArrayList<>(mStore.entrySet()));
     }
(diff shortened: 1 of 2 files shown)


thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterScope.java
```diff
@@ -54,13 +58,19 @@ public boolean isEmpty() {
...
-    public TiPresenter remove(final String id) {
+    public TiPresenter remove(@NonNull final String id) {
         final TiPresenter presenter = mStore.remove(id);
         TiLog.d(TAG, "remove " + id + " " + presenter);
         return presenter;
     }
 
-    public void save(final String id, final TiPresenter presenter) {
+    public void save(@NonNull final String id, @NonNull final TiPresenter presenter) {
+        if (id == null) {
```
StefMa: I don't know if I agree with that 🤔 
According to the `@NonNull` doc:
```
 Denotes that a parameter, field or method return value can never be null.
```

Which means. If we put `null` in here, we do something wrong.
Of course, such issues  may happen with mutation testing ( 😉  ) but normally it doesn't by a human (if we add the annotation)....

... 🤔

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/PresenterScopeTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+    @Test
+    public void addTwoMappings() throws Exception {
+
+        final PresenterScope scope = new PresenterScope();
+        final TiPresenter presenter1 = new TiPresenter() {
+        };
+        scope.save("a", presenter1);
+
+        // override with same id
```
StefMa: That is wrong ;)

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/PresenterScopeTest.java
```diff
@@ -0,0 +1,147 @@
...
+        final TiPresenter presenter2 = new TiPresenter() {
```
StefMa: We don't need to create another presenter here.. 
Just reuse presenter1

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterScope.java
passsy: Then we should also remove other checks like those in [`TiPresenter`](https://github.com/grandcentrix/ThirtyInch/blob/master/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java#L158)

For me the `@NonNull` annotation is just a warning: "Don't put null here, it won't work".
The problem here is that it will work in a way which is not intended. That's why I want to explicitly disallow this option by throwing the Exception. The annotation is just an additional hint which helps devs to get a warning before running into the runtime exceptions.

Btw, Kotlin does the same. When a parameter is non nullable the compile adds a non null check at the beginning of the method in byte code.

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/PresenterScopeTest.java
passsy: Oh that's a good edge case. We should prevent adding a presenter twice with a different id.

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/internal/PresenterScope.java
```diff
@@ -54,13 +58,34 @@ public boolean isEmpty() {
...
+        if (mStore.get(id) != null) {
+            throw new IllegalStateException("There is already a presenter saved with id "
+                    + id + " " + presenter);
+        }
+
+        // saving a presenter twice with a different id is not supported
+        for (final Map.Entry<String, TiPresenter> entry : mStore.entrySet()) {
+            if (entry.getValue().equals(presenter)) {
+                throw new IllegalStateException("Presenter is already saved with different id "
+                        + entry.getKey() + " " + presenter);
```
passsy: missing test

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/internal/PresenterScopeTest.java
```diff
@@ -0,0 +1,166 @@
...
+            scope.save(null, new TiPresenter() {
+            });
+            fail("did not throw");
+        } catch (IllegalStateException e) {
+            assertThat(e).hasMessageContaining("null").hasMessageContaining("id");
+        }
+    }
+
+    @Test
+    public void saveSamePresenterTwiceThrows() throws Exception {
```
StefMa: Duplicated test. This is the same like `overrideMappingThrows`, right? 🤔

passsy: That's the reason why `Adding the presenter twice with different id` was missing. The implementation was wrong and therefore duplicated.
Fixed it :)

PR #120, opened 2017-10-01
Merge test

### LINKED ISSUE #118: Merge thirtyinch-test in thirtyinch
Currently we have a extra `thirtyinch-test` dependency which includes just [a single file](https://github.com/grandcentrix/ThirtyInch/blob/master/test/src/main/java/net/grandcentrix/thirtyinch/test/TiPresenterInstructor.java).
Because it is way more overhead to maintain these extra package we agreed that we want to merge it into the "[main](https://github.com/grandcentrix/ThirtyInch/tree/master/thirtyinch)" `thirtyinch` module.

We want to have a similar idea like RxJava in tests.
Just calling: 
```kotlin
val presenterInstructor = MyPresenterImpl().test()
```
Currently we are not really sure if we really want to return the `PresenterInstructor` or another (helper) object which hold the real presenter and the instructor...
Maybe later we want to add more functionality to it which can nicely delegate the "actions" (or maybe `asserts` 🤔 ) to the correct object...

StefMa:
This will fix #118 

Currently I've just copied the Instructor code into the `TiTestPresenter`, added some tests (why the TestPresener is needed - not testing it themself) and added a `test()` method...

### DIFF
--- a/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
+++ b/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
@@ -16,8 +16,6 @@
 package net.grandcentrix.thirtyinch;
 
 
-import net.grandcentrix.thirtyinch.internal.OneTimeRemovable;
-
 import android.app.Activity;
 import android.content.Intent;
 import android.support.annotation.NonNull;
@@ -25,12 +23,13 @@ import android.support.annotation.Nullable;
 import android.support.annotation.VisibleForTesting;
 import android.support.v4.app.Fragment;
 import android.widget.Toast;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Queue;
 import java.util.concurrent.Executor;
 import java.util.concurrent.LinkedBlockingQueue;
+import net.grandcentrix.thirtyinch.internal.OneTimeRemovable;
+import net.grandcentrix.thirtyinch.test.TiTestPresenter;
 
 /**
  * Represents the Presenter of the popular Model-View-Presenter design pattern. If used with {@link
@@ -144,7 +143,6 @@ public abstract class TiPresenter<V extends TiView> {
             }
         };
     }
-
 
     /**
      * bind a new view to this presenter.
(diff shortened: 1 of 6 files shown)


thirtyinch/src/test/java/net/grandcentrix/thirtyinch/test/TiTestPresenterTest.java
```diff
@@ -0,0 +1,68 @@
...
+                sendToView(new ViewAction<MockTiView>() {
+                    @Override
+                    public void call(MockTiView tiView) {
+                        tiView.helloWorld();
+                    }
+                });
+            }
+        };
+        mockTiPresenter.create();
+        mockTiView = new MockTiView() {
```
passsy: use `Mockito.mock(MockTiView.class);` and verify that `helloWorld()` was called.

StefMa:
For me the changes look good.
I think we can ignore codecov because of -0,02% 🙄 

Anything missing here? Otherwise I would say :shipit:

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java
```diff
@@ -386,6 +386,10 @@ public void setUiThreadExecutor(@Nullable final Executor uiThreadExecutor) {
         mUiThreadExecutor = uiThreadExecutor;
     }
 
+    public TiTestPresenter<V> test() {
```
passsy: Thought about this method a lot, especially if it is extensible enough for the future. It is. When my presenter requires additional setup for testing one can override this method. That's great 🚢

passsy: Missing documentation for `test()`. Copy from `TiTestPresenter`

thirtyinch/src/main/java/net/grandcentrix/thirtyinch/test/TiTestPresenter.java
```diff
@@ -0,0 +1,137 @@
...
+        switch (state) {
+            case INITIALIZED:
+                mPresenter.create();
+                break;
+            case VIEW_DETACHED:
+                // already there
+                break;
+            case VIEW_ATTACHED:
+                mPresenter.detachView();
+                mPresenter.setUiThreadExecutor(null);
```
passsy: task for Hacktoberfest: Add tests for `TiTestPresenter`

PR #123, opened 2017-10-02
Add project codestyles

### LINKED ISSUE #122: Put codestyle into ThirtyInch
Since we [open sourced](https://github.com/grandcentrix/androidcodestyle) our code style I think about to add it here.
We can add it together with a CONTRIBUTING.md ( 🤔 ) to Ti. We won't have conflicts or "rerange the code everytime" if third parties contribute anymore...

What do you think @passsy @rberghegger ❓

rberghegger:
This will fix #122 
As discussed in #122 this pull request adds project specific code styles based on the grandcentrix android code styles (https://github.com/grandcentrix/androidcodestyle) to the project.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -1,0 +1,775 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectCodeStyleSettingsManager">
+    <option name="PER_PROJECT_SETTINGS">
+      <value>
+        <option name="FIELD_NAME_PREFIX" value="m" />
+        <option name="STATIC_FIELD_NAME_PREFIX" value="m" />
+        <option name="GENERATE_FINAL_LOCALS" value="true" />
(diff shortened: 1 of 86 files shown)


passsy:
Not sure if we should convert all sources within this PR or not 🤔

rberghegger:
@passsy Good point. Was thinking about that as well. 🤔  But I wanted to wait for your and @StefMa opinions before actually applying those styles.

StefMa:
Hmm.. Since we don't have that much PR I would suggest to do it. 👍

passsy:
true, just do it. If there are conflicts we can solve them by applying the code style as well

rberghegger:
Done. I've applied the new project codestyle to every module individually. Additionally I've added `CONTRIBUTING.md` which contains a few informations regarding the codestyle.

StefMa:
I would suggest to remove the TK from the contribution file. Also start with a single # and use two for code style... 👍 rest lgtm

Issue #121, opened 2017-10-01
Use assertJ everywhere


StefMa:
Currently we have "hundreds" of different "test matchers".
* The "build in" one `assertEquals("ThirtyInch", "ThirtyInch")`
* Hamcrest `assertThat("ThirtyInch", isEqualTo("ThirtyInch"))`
* assertJ `assertThat("ThirtyInch").isEqualTo("ThirtyInch")`

Each test look different. Which is bad.

We should use assertJ everywhere and get rid of the other matcher libs.
To prevent using Hamcrest in the future it would be great if we can exclude it from... Don't know which dependency has hamcrest as dependency. But we have to found out it and exclude it form them if possible.

passsy:
This is a very good task for Hacktoberfest!
I'm not sure if we can exclude it. You can get the dependnecies tree with `./gradlew :thirtyinch:dependnecies`

sumedh123:
@StefMa We just need to replace 'assertEquals' and 'assertThat' by 'assertJ' right? Or anything else needs to be done?

StefMa:
Exactly. But assertJ isn't a method 🙃. From both, hamcrest and assertJ, the matches are assertThat

vpondala:
I am working on this.

StefMa:
#128 has merged and we are now using assertJ everywhere.
But it hasn't removed hamcrest as dependency.

It is still open which dependency brings hamcrest build in.

passsy:
I tried to remove it
```gradle
    testCompile ("junit:junit:$junitVersion") {
        exclude group: 'org.hamcrest', module: 'hamcrest-core'
    }
```

Results in 
```
java.lang.NoClassDefFoundError: org/hamcrest/SelfDescribing
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:760)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:368)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:362)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:361)
[...]

PR #128, opened 2017-10-07
#121: Use assertJ everywhere

vpondala:
Migrated all assertions to AssertJ.

### DIFF
--- a/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/SendToViewTest.java
+++ b/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/SendToViewTest.java
@@ -16,7 +16,7 @@
 package net.grandcentrix.thirtyinch;
 
 
-import static org.assertj.core.api.Java6Assertions.*;
+import static org.assertj.core.api.Assertions.*;
 import static org.mockito.Mockito.*;
 
 import android.support.annotation.NonNull;
--- a/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/SendToViewTest.java
+++ b/thirtyinch/src/test/java/net/grandcentrix/thirtyinch/SendToViewTest.java
@@ -16,8 +16,6 @@
 package net.grandcentrix.thirtyinch;
 
 
-import static junit.framework.Assert.assertNotSame;
-import static junit.framework.Assert.assertTrue;
 import static org.assertj.core.api.Java6Assertions.*;
 import static org.mockito.Mockito.*;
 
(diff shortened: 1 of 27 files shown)


vpondala:
My local build succeeded. This is my 1st Android build, not sure how to fix it, Can you please guide?

StefMa:
Hey, thank you for your PR!

Sure.
Try googeling for "Duplicate files copied in APK META-INF/License".

The problem is that two or more deoencies try to add these file into the
apk. Which is not allowed.
I am not  hundred percent sure but it seems that the assertJ and the
support test runner wants to include that file.

Just try to exclude it and the thirtyinch module  and it should work 👍

vpondala:
I added packagingOptions { exclude 'META-INF/LICENSE' } and the build passes now. Please have a look at the diffs and let me know if the changes look good.

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/SendToViewTest.java
```diff
@@ -118,9 +116,10 @@ public Thread newThread(final Runnable r) {
             public void call(final TiView tiView) {
                 // Then the work gets executed on the ui thread
                 final Thread currentThread = Thread.currentThread();
-                assertNotSame(testThread, currentThread);
-                assertTrue("executed on wrong thread",
-                        "test ui thread".equals(currentThread.getName()));
+                assertThat(testThread).isNotSameAs(currentThread);
+                assertThat("test ui thread")
+                        .as("executed on wrong thread")
```
passsy: didn't know this one, great 🎉

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiLifecycleObserverTest.java
```diff
@@ -303,14 +302,14 @@ public void onChange(final TiPresenter.State state,
         mPresenter.detachView();
 
         final Object[] beforeLast = states.get(states.size() - 2);
-        assertEquals(beforeLast[0], TiPresenter.State.VIEW_DETACHED);
-        assertEquals(beforeLast[1], false);
-        assertNotNull(beforeLast[2]);
+        assertThat(beforeLast[0]).isEqualTo(TiPresenter.State.VIEW_DETACHED);
+        assertThat(beforeLast[1]).isEqualTo(false);
```
passsy: better use `.isFalse()`

passsy:
There's a mixup between `import static org.assertj.core.api.Assertions.*;` and `import static org.assertj.core.api.Java6Assertions.*;`. The `Java6Assertions` was required for androidTests in the past. I'm not sure if this is still the case, AssertJ improved Android support a lot recently.

Try to use `import static org.assertj.core.api.Assertions.*;` everywhere

thirtyinch/src/test/java/net/grandcentrix/thirtyinch/TiLifecycleObserverTest.java
vpondala: In these cases, it's AbstractObjectAssert, which does not have isTrue()/isFalse().

passsy:
Thanks for your contribution!

PR #135, opened 2017-11-14
Update to gradle plugin 3.0

StefMa:
With that change we can think about our dependencies again.

Do we want to "bundle" the dependencies with our libraries or not?
Which means, do we want to use `api` or `implementation`?
If we decided we use `implementation` we can also think about to use `compileOnly` for "all", right?
Since we don't really need a special version or something. Our implementations are very "basic" and I think all future releases of our dependencies will support our implementation...

What do you think?

> **Note*:* While upgrading I had some issues with jacoco. So i declared the jacoco version and added the test coverage in the sample. [jacoco closure is depcreated ](https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.BaseExtension.html#com.android.build.gradle.BaseExtension:jacoco(org.gradle.api.Action)). We have to add it to the classpath...

### DIFF
--- a/thirtyinch-rx/build.gradle
+++ b/thirtyinch-rx/build.gradle
@@ -36,14 +36,14 @@ android {
 }
 
 dependencies {
-    compile project(':thirtyinch')
-    compile 'io.reactivex:rxjava:1.3.2'
-    compile 'com.artemzin.rxjava:proguard-rules:1.2.7.0'
-    provided "com.android.support:support-annotations:$supportLibraryVersion"
+    implementation project(':thirtyinch')
+    implementation 'io.reactivex:rxjava:1.3.2'
+    implementation 'com.artemzin.rxjava:proguard-rules:1.2.7.0'
+    compileOnly "com.android.support:support-annotations:$supportLibraryVersion"
 
-    testCompile "junit:junit:$junitVersion"
-    testCompile "org.mockito:mockito-core:$mockitoVersion"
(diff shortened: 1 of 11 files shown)


thirtyinch-rx/build.gradle
```diff
@@ -36,14 +36,15 @@ android {
 }
 
 dependencies {
-    compile project(':thirtyinch')
-    compile 'io.reactivex:rxjava:1.3.2'
-    compile 'com.artemzin.rxjava:proguard-rules:1.2.7.0'
-    provided "com.android.support:support-annotations:$supportLibraryVersion"
+    implementation project(':thirtyinch')
+    implementation 'io.reactivex:rxjava:1.3.2'
```
passsy: rx should be `api`

thirtyinch-rx2/build.gradle
```diff
@@ -36,13 +36,13 @@ android {
 }
 
 dependencies {
-    compile project(':thirtyinch')
-    compile 'io.reactivex.rxjava2:rxjava:2.1.4'
-    provided "com.android.support:support-annotations:$supportLibraryVersion"
+    implementation project(':thirtyinch')
+    implementation 'io.reactivex.rxjava2:rxjava:2.1.4'
```
passsy: same here -> `api`

StefMa:
We decided to use #136

Issue #34, opened 2016-11-02
deliverLatestCacheToView delivers all previously emitted items


gborland:
`RxTiPresenterUtils.deliverLatestCacheToView` caches and re-delivers **all** items emitted by the source.

Look at this log output. When the view is detached, several items are emitted by the source, but not delivered. This is correct so far.

After the view is reattached, all of the items are delivered again, inclusing those which had already been delivered (`c9915aa`) and those which should have been dropped (`6b60904`).

```
11-02 12:45:45.392 D: Emitting c9915aa
11-02 12:45:45.421 I: Delivering c9915aa
11-02 12:46:02.199 D: Emitting 75d8118
11-02 12:46:02.210 I: Delivering 75d8118
11-02 12:46:02.211 D: Emitting a0e7956
11-02 12:46:02.255 I: Delivering a0e7956

VIEW DETACHED

11-02 12:46:44.932 D: Emitting 6b60904
11-02 12:46:45.052 D: Emitting 27a3122
11-02 12:46:45.668 D: Emitting 50cb170
11-02 12:46:45.682 D: Emitting 1ae3d6e
11-02 12:46:46.478 D: Emitting efbdc0f
11-02 12:46:46.493 D: Emitting b085ea5

VIEW REATTACHED

11-02 12:47:09.806 I: Delivering c9915aa
11-02 12:47:09.807 I: Delivering 75d8118
11-02 12:47:09.807 I: Delivering a0e7956
11-02 12:47:09.808 I: Delivering 6b60904
11-02 12:47:09.809 I: Delivering 27a3122
11-02 12:47:09.809 I: Delivering 50cb170
11-02 12:47:09.810 I: Delivering 1ae3d6e
11-02 12:47:09.811 I: Delivering efbdc0f
11-02 12:47:09.812 I: Delivering b085ea5
11-02 12:47:15.458 D: Emitting 4cdffd
11-02 12:47:15.459 I: Delivering 4cdffd
```

Here is the code:
```kotlin
val subscription =
        Observable.just(/* some object */)
                .doOnNext { c -> Timber.d("Emitting %x".format(c.hashCode())) }
                .delaySubscription(5, TimeUnit.SECONDS)
                .subscribeOn(Schedulers.io())
                .observeOn(AndroidSchedulers.mainThread())
                .compose(RxTiPresenterUtils.deliverLatestCacheToView<Object>(this))
                .subscribe { c ->
[...]

PR #137, opened 2017-11-30
Add deliverLatestToView to rx2/RxTiPresenterUtils

GrahamBorland:
This adds an implementation of `deliverLatestToView` to rx2/RxTiPresenterUtils.

It uses only existing RxJava2 operators, rather than the `OperatorSemaphore` from the Rx1 utils. I did initially try to port `OperatorSemaphore` but it quickly became terrifying.

I have not implemented the other two (`deliverLatestCacheToView` and `deliverToView`) since I have no use for them yet, and I'm not really sure what they are meant to do anyway. (Their documentation is confusing - see #34).

Partial fix for #9.

### DIFF
--- a/thirtyinch-rx2/src/main/java/net/grandcentrix/thirtyinch/rx2/RxTiPresenterUtils.java
+++ b/thirtyinch-rx2/src/main/java/net/grandcentrix/thirtyinch/rx2/RxTiPresenterUtils.java
@@ -87,9 +87,9 @@ public class RxTiPresenterUtils {
     }
 
     /**
-     * Returns a transformer that will delay onNext, onError and onComplete emissions unless a view
-     * become available. getView() is guaranteed to be != null during all emissions. This
-     * transformer can only be used on application's main thread.
+     * Returns a transformer that will delay onNext, onError and onComplete emissions until a view
+     * become available. getView() is guaranteed to be != null during all emissions, provided that this
+     * transformer is only used on the application's main thread.
      * <p/>
      * If this transformer receives a next value while the previous value has not been delivered,
      * the previous value will be dropped.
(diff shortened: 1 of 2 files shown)


thirtyinch-rx2/src/main/java/net/grandcentrix/thirtyinch/rx2/RxTiPresenterUtils.java
```diff
@@ -66,4 +86,54 @@ public boolean isDisposed() {
         }).distinctUntilChanged();
     }
 
+    /**
+     * Returns a transformer that will delay onNext, onError and onComplete emissions unless a view
+     * become available. getView() is guaranteed to be != null during all emissions. This
+     * transformer can only be used on application's main thread.
```
passsy: I must admit, when I wrote the `rx1` version I [copied](https://github.com/konmik/nucleus/blob/4167820fbb4061d8a6f303de3903751a8e136793/nucleus/src/main/java/nucleus/presenter/RxPresenter.java#L182) most of the docs.
Actually the main thread restriction is not true it's only recommended to make sure the view is always non null. Attach/Detach happens on the main thread. When code executes on any other thread it's not guaranteed that those actions happen before the view attach/detach state changes. So it may happen that the emission of `T` happens after the view got detached.

GrahamBorland: I'm happy to rewrite the comment. The behaviour should be clear from the tests anyway, I hope.

thirtyinch-rx2/src/test/java/net/grandcentrix/thirtyinch/rx2/RxTiPresenterUtilsTest.java
```diff
@@ -57,4 +59,116 @@ public void testIsViewReady_DisposeBeforeAttachView_ShouldRemoveCallback() throw
...
+        mPresenter.attachView(mView);
+        source.onNext(6);
+
+        testObserver.assertNotComplete();
+        testObserver.assertNoErrors();
+        testObserver.assertValuesOnly(2, 3, 5, 6);
+    }
+
+    @Test
+    public void testDeliverLatestToView_SingleItemViewComesAndGoes() throws Exception {
```
passsy: We should add this test to rx1 as well. I'm actually not sure this is working.

StefMa: For that I've created a new issue. See #138

GrahamBorland:
Comment adjusted, thanks.

PR #142, opened 2018-02-20
Made sendToView public to allow to override or extend the function.

rehlma1989:


StefMa:
You could extend or override the function with the protection modifier as
well.

What is your use case? 🤔

jreehuis:
@rehlma1989 This method is intended just for Presenter internal use. So it should not be opened to external classes.
And as @StefMa said, in the actual Presenter it is overridable anyways.

passsy:
The protection modifier prevents us to create an extension function in Kotlin which uses `sendToView`

StefMa:
To clean up some things:
The goal idea was to create a kotlin extension like:
```kotlin
fun <V : TiView> TiPresenter<V>.sendToViewKotiln(block: (V.() -> Unit)) {
    sendToView { v -> v.block() }
}
```

This will give us the benefit that the receiver is not a param (`it`) but `this`.
Which means instead of having this:
```kotlin
fun aMethod() {
   sendToView { it.showLoading(); it.showAnotherThings(); it.doWhatever() }
}
```
we can use:
```kotlin
fun aMethod() {
   sendToView { showLoading(); showAnotherThings(); doWhatever() }
}
```

Since Kotlin extensions are just static java methods we can't create such a extension without a public method. `sendToView` is protected and we haven't access to it from a static method.
That is the reason why this PR was opened...

StefMa:
To speak about pros/cons:

After seeing the benefits of it I'm not against this change.
Normally the presenter is only available/accessable via a View (Activtiy/Fragment). And I think no one will ever use `presenter.sendToView {}` from a View...

jreehuis:
That UseCase definitely make sense. Is it possible to restrict the usage than by `@RestrictTo(RestrictTo.Scope.SUBCLASSES)` or another scope?

passsy:
`RestrictTo(SUBCLASSES)` sound great!

rehlma1989:
Great! Just did it.

PR #150, opened 2018-07-06
Basic implementation for TI kotlin

StefMa:
This will introduce a "basic" implementation for Ti-Kotlin.

Currently it is just the module and the extension function 
`TiPresenter.sendToViewKotlin(block: View.() -> Unit)`.
See also this PR: https://github.com/grandcentrix/ThirtyInch/pull/142

I'm not quite happy with the function name. But because Kotlin uses the implementation function over the extension function it isn't possible to name them equally.
If you have a better name feel free to post it here...


thirtyinch-kotlin/src/main/java/net/grandcentrix/thirtyinch/kotlin/TiPresenter.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Example:
+ * ```
+ * // Before
+ * presenter.sendToView { it.aViewMethod() }
+ * // After
+ * presenter.sendToViewKotlin { aViewMethod() }
+ * ```
+ */
+@SuppressLint("RestrictedApi")
+fun <V : TiView> TiPresenter<V>.sendToViewKotlin(block: V.() -> Unit) = sendToView { block.invoke(it) }
```
Syex: How about at least name it `sendToViewKt` as all Kotlin stuff is suffixed with `Kt`?

StefMa: Thanks for the recommendation.
I think we should go with @Syex idea here because... well, it make sense 🙃 

Using a complete different name for an "extension" is just confusing IMO.

jreehuis: I would prefer to go with @passsy 's suggestion. We should not include some language specific abbreviations in our public API.
My prefered candidate would be `withView`. Reason: It makes sense.

StefMa:
🔴 Ready for another round 👍 
Thanks @passsy Great suggestion

README.md
```diff
@@ -250,6 +253,30 @@ public class HelloWorldActivity extends TiActivity<HelloWorldPresenter, HelloWor
 
 `LoggingInterceptor` is available as module and logs all calls to the view.
 
+### Kotlin
+
+Using Kotlin these days is a no brainer.
```
jreehuis: Consider adding a part that the user has to include `net.grandcentrix.thirtyinch:thirtyinch-kotlin`

README.md
```diff
@@ -250,6 +253,30 @@ public class HelloWorldActivity extends TiActivity<HelloWorldPresenter, HelloWor
...
+`ThirtyInch` provides some extension methods to improve itself even further!  
+
+Take a look at the `sendToViewKt` example:
```
jreehuis: change `sendToViewKt ` here to `deliverToView` aswell

thirtyinch-kotlin/src/main/java/net/grandcentrix/thirtyinch/kotlin/TiPresenter.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Example:
+ * ```
+ * // Before
+ * presenter.sendToView { it.aViewMethod() }
+ * // After
+ * presenter.deliverToView { aViewMethod() }
+ * ```
+ */
+@SuppressLint("RestrictedApi")
+fun <V : TiView> TiPresenter<V>.deliverToView(block: V.() -> Unit) = sendToView { block.invoke(it) }
```
jreehuis: Consider to write `block(it)` instead of `block.invoke(it)`

jreehuis:
Nice addition @StefMa ! 🐶

PR #151, opened 2018-07-10
Android Studio 3.1 update

passsy:
Upgrade `com.android.tools.build:gradle` to `3.1.3`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ android:
     - tools
 
     # The BuildTools version used by your project
-    - build-tools-26.0.2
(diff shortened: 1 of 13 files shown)


StefMa:
I blocked it because of #152 .

As I remind me correctly there was an issue with the Plugin and the "old" novoda/bintray-release.
Anyway. We could wait until #152 is merged and then review/merge this one 👍

passsy:
Just tested releasing to mavenLocal as well as uploading with `dryRun=true`. Both works without #152, the `novoda/bintray-release` plugin works.

PR #152, opened 2018-07-10
New bintray release

StefMa:
This will use the new BintrayRelease plugin (https://github.com/StefMa/bintray-release/pull/1)
It's not jet available at jcenter.

I've also removed the bintray-release.gradle file because it felt wrong somehow...

## How to test:
Try to publish the project.
Either with `-PdryRun=true` or with `-PdryRun=false` but to a different orga than grandcentrix


passsy:
I'm definitely interested in the new publishing plugin but due to planned changes in the plugin I'll put this PR on hold.

What I don't like is that `version` and `group` are defined outside of the extension. The version should be read from `android.defaultConfig.version` (if possibel) and the I'd prefer defining the `group` inside the closure.

I tried to rebuild the `bintrayRelease.gradle` to reduce duplicated code without success. I got this *strange* error when applying the plugin.
```
Extension of type 'LibraryExtension' does not exist. Currently registered extension types:
 Currently registered extension types: [ExtraPropertiesExtension, DefaultArtifactPublicationSet, ReportingExtension, NamedDomainObjectContainer<BaseVariantOutput>, LibraryExtension, AndroidArtifactsExtension, PublishingExtension]
```

StefMa:
Its working.
I've also checked that *all* POM files are identical with the one from 0.9.0.

PR #158, opened 2018-08-27
Feature: Lint Checks

### LINKED ISSUE #100: ThirtyInch Lint Checks

jreehuis:
#### Description
This implementation adopts and updates #111.
Fixes #100 

The updates include:
- Using the now stable Lint version to use UAST
- Bundles the Lint checks with TI
- Fixing the detection scope so Android Studio will report the error in the editor
- Creating an own TI Lint Category
- `TiIssue` is now a sealed class
- some Kotlin tweaks

#### How to Test
Play with the following scenarios:
- Remove the `implements HelloWorldView` in the `HelloWorldActivity` of the sample app
- Remove the `implements HelloWorldView` but override `provideView()` in the `HelloWorldActivity` of the sample app
- Use TI v`0.9.1-SNAPSHOT` of this branch in another project and have an TiActivity which does not implement the given `TiView` subclass

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,4 +45,5 @@ ext {
     assertjVersion = '2.8.0'
     supportTestVersion = '1.0.1'
     espressoVersion = '3.0.1'
+    lintVersion = '26.1.2'
 }
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:3.1.3"
+        classpath "com.android.tools.build:gradle:3.1.3" // if you update this, also update the lintVersion below
         classpath "com.vanniktech:gradle-android-junit-jacoco-plugin:0.10.0"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
@@ -45,5 +45,9 @@ ext {
     assertjVersion = '2.8.0'
     supportTestVersion = '1.0.1'
     espressoVersion = '3.0.1'
-    lintVersion = '26.1.2'
+
+    // According to https://github.com/googlesamples/android-custom-lint-rules/tree/master/android-studio-3
+    // the lint version should match to the used Android Gradle Plugin by the formula "AGP Version X.Y.Z + 23.0.0"
+    // E.g. "AGP Version 3.1.3 + 23.0.0 = Lint Version 26.1.3"
+    lintVersion = '26.1.3'
 }
--- a/thirtyinch-lint/README.md
+++ b/thirtyinch-lint/README.md
@@ -1,0 +1,1 @@
+# ThirtyInch - Lint
--- a/thirtyinch-lint/README.md
+++ b/thirtyinch-lint/README.md
@@ -1,1 +1,0 @@
-# ThirtyInch - Lint
--- a/thirtyinch-lint/build.gradle
+++ b/thirtyinch-lint/build.gradle
@@ -17,7 +17,6 @@ dependencies {
     compileOnly "com.android.tools.lint:lint-api:$lintVersion"
     compileOnly "com.android.tools.lint:lint-checks:$lintVersion"
 
-    testImplementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
     testImplementation 'junit:junit:4.12'
     testImplementation "org.assertj:assertj-core:$assertjVersion"
 
--- a/thirtyinch-lint/build.gradle
+++ b/thirtyinch-lint/build.gradle
@@ -31,6 +31,8 @@ jar {
     manifest {
         attributes("Manifest-Version": 1.0)
         attributes("Lint-Registry": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
+        // The TI checks are build with the new 3.0 APIs (including UAST) so we should also register the v2 lint registry.
+        attributes("Lint-Registry-v2": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
     }
 }
 
--- a/thirtyinch-lint/build.gradle
+++ b/thirtyinch-lint/build.gradle
@@ -7,6 +7,10 @@ repositories {
     jcenter()
 }
 
+configurations {
+    lintChecks
+}
+
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
 
@@ -19,10 +23,13 @@ dependencies {
 
     testImplementation "com.android.tools.lint:lint:$lintVersion"
     testImplementation "com.android.tools.lint:lint-tests:$lintVersion"
+
+    lintChecks files(jar)
 }
 
 jar {
     manifest {
+        attributes("Manifest-Version": 1.0)
         attributes("Lint-Registry": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
     }
 }
--- a/thirtyinch-lint/build.gradle
+++ b/thirtyinch-lint/build.gradle
@@ -23,6 +23,7 @@ dependencies {
 
 jar {
     manifest {
+        attributes("Lint-Registry": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
     }
 }
 
--- a/thirtyinch-lint/build.gradle
+++ b/thirtyinch-lint/build.gradle
@@ -35,8 +35,6 @@ jar {
     }
 }
 
-sourceCompatibility = 1.6
-
 compileKotlin {
     kotlinOptions {
         jvmTarget = "1.8"
--- a/thirtyinch-lint/build.gradle
+++ b/thirtyinch-lint/build.gradle
@@ -37,7 +37,6 @@ jar {
 
 sourceCompatibility = 1.6
 
-defaultTasks 'assemble'
 compileKotlin {
     kotlinOptions {
         jvmTarget = "1.8"
--- a/thirtyinch-lint/build.gradle
+++ b/thirtyinch-lint/build.gradle
@@ -1,0 +1,42 @@
+plugins {
+    id "org.jetbrains.kotlin.jvm"
+    id "jacoco"
+}
+
+repositories {
+    jcenter()
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+
+    compileOnly "com.android.tools.lint:lint-api:$lintVersion"
+    compileOnly "com.android.tools.lint:lint-checks:$lintVersion"
+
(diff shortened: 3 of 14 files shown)


build.gradle
```diff
@@ -45,4 +45,9 @@ ext {
     assertjVersion = '2.8.0'
     supportTestVersion = '1.0.1'
     espressoVersion = '3.0.1'
+
+    // According to https://github.com/googlesamples/android-custom-lint-rules/tree/master/android-studio-3
+    // the lint version should match to the used Android Gradle Plugin by the formula "AGP Version X.Y.Z + 23.0.0"
+    // E.g. "AGP Version 3.1.3 + 23.0.0 = Lint Version 26.1.3"
+    lintVersion = '26.1.3'
```
StefMa: We could also calculate this by ourselfs, right? 🤔 
Just extract the AGP version and "add" the values to 23.0.0 ...

thirtyinch-lint/README.md
```diff
@@ -0,0 +1 @@
+# ThirtyInch - Lint
```
StefMa: No one of our modules have its own README.
Even if this brings no value I prefer to remove it

thirtyinch-lint/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+    lintChecks
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+
+    compileOnly "com.android.tools.lint:lint-api:$lintVersion"
+    compileOnly "com.android.tools.lint:lint-checks:$lintVersion"
+
+    testImplementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
```
StefMa: This dependency is already implement as a `implementation` dependency.
Because `testImplementation` extends `implementation` there is no need to declare it again

thirtyinch-lint/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+        attributes("Manifest-Version": 1.0)
+        attributes("Lint-Registry": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
+        // The TI checks are build with the new 3.0 APIs (including UAST) so we should also register the v2 lint registry.
+        attributes("Lint-Registry-v2": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
+    }
+}
+
+sourceCompatibility = 1.6
+
+defaultTasks 'assemble'
```
StefMa: Is that really needed? 🤔 
I prefer to remove it.
Running gradle without any specific task is not that common...

thirtyinch-lint/build.gradle
```diff
@@ -0,0 +1,52 @@
...
+jar {
+    manifest {
+        attributes("Manifest-Version": 1.0)
+        attributes("Lint-Registry": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
+        // The TI checks are build with the new 3.0 APIs (including UAST) so we should also register the v2 lint registry.
+        attributes("Lint-Registry-v2": "net.grandcentrix.thirtyinch.lint.TiLintRegistry")
+    }
+}
+
+sourceCompatibility = 1.6
```
StefMa: Why do we need this?

thirtyinch-lint/src/main/kotlin/net/grandcentrix/thirtyinch/lint/detector/MissingViewInCompositeDetector.kt
```diff
@@ -0,0 +1,89 @@
...
+
+private const val ADD_PLUGIN_METHOD = "addPlugin"
+private const val TI_ACTIVITY_PLUGIN_NAME = "TiActivityPlugin"
+private const val TI_FRAGMENT_PLUGIN_NAME = "TiFragmentPlugin"
+private val CA_CLASS_NAMES = listOf(
+        "com.pascalwelsch.compositeandroid.activity.CompositeActivity",
+        "com.pascalwelsch.compositeandroid.fragment.CompositeFragment"
+)
+
+class MissingViewInCompositeDetector : BaseMissingViewDetector() {
```
StefMa: Don't know if it make sense because we wanted to drop the support vor CompositeAndroid anyway, right @passsy?
But as long as it is implemented now and work we could leave it like it is...

thirtyinch-lint/src/main/kotlin/net/grandcentrix/thirtyinch/lint/detector/MissingViewInThirtyInchDetector.kt
```diff
@@ -0,0 +1,57 @@
...
+                MissingViewInThirtyInchDetector::class.java,
+                "When using ThirtyInch, a class extending TiActivity, TiFragment or CompositeActivity " +
+                        "has to implement the TiView interface associated with it in its signature, " +
+                        "or implement `provideView()` instead to override this default behaviour."
+        )
+    }
+
+    override fun applicableSuperClasses() = TI_CLASS_NAMES
+
+    override val issue: Issue; get() = ISSUE
```
StefMa: Why is here an semicolon?
Should be removed. See also [this style guide](https://kotlinlang.org/docs/reference/coding-conventions.html#property-formatting).

thirtyinch-lint/src/main/kotlin/net/grandcentrix/thirtyinch/lint/detector/MissingViewInThirtyInchDetector.kt
```diff
@@ -0,0 +1,57 @@
...
+private val TI_CLASS_NAMES = listOf(
+        "net.grandcentrix.thirtyinch.TiActivity",
+        "net.grandcentrix.thirtyinch.TiFragment"
+)
+
+class MissingViewInThirtyInchDetector : BaseMissingViewDetector() {
+    companion object {
+        val ISSUE = MissingView.asLintIssue(
+                MissingViewInThirtyInchDetector::class.java,
+                "When using ThirtyInch, a class extending TiActivity, TiFragment or CompositeActivity " +
```
StefMa: This issue don't catch the issues for the CompositeActivity.
Please adjust the comment

thirtyinch-lint/src/main/kotlin/net/grandcentrix/thirtyinch/lint/detector/MissingViewInThirtyInchDetector.kt
jreehuis: This semicolon is needed to keep it all in one line. But I removed the getter overriding completely.

thirtyinch-lint/src/main/kotlin/net/grandcentrix/thirtyinch/lint/detector/MissingViewInCompositeDetector.kt
jreehuis: Correct, as long it is still present we should support it.

build.gradle
jreehuis: Yes, we can calculate this ourself. But I would prefer to change that still manually:
- It is stated "should" and so the actual version might differ from the formula
- Google recently just decoupled the Support Lib Versions and so it might be that this version gets uncoupled from the build tools version aswell

jreehuis:
@StefMa I addressed all your comments. Would be great if you could have another look on this PR. :)

StefMa:
Hey, sure.
After I am back from my sick leave I will take a look into it 👍

StefMa:
Please update the README how to add this to a project

StefMa:
Works also with Java 🤘 
bildschirmfoto 2018-09-17 um 10 31 16
bildschirmfoto 2018-09-17 um 10 31 46

StefMa:
Something like this shouldn't be an issue:
bildschirmfoto 2018-09-17 um 10 33 05

thirtyinch/build.gradle
```diff
@@ -49,3 +55,16 @@ configurePublish {
     artifactId = 'thirtyinch'
     desc = 'a Model View Presenter library for Android'
 }
+
+task copyLintJar(type: Copy) {
+    from(configurations.lintChecks) {
+        rename { String fileName -> 'lint.jar' }
+    }
+    into 'build/intermediates/lint/'
```
StefMa: Is it common to move the lint check into the final AAR? 🤔 
What about to publish a separate jar/artifact and make this optional? 🤔

StefMa:
Other than that it looks great 🎉 
Thanks @jreehuis

jreehuis:
Thanks for the extensive tests! 😃 
Especially the abstract class scenario will get complicated... 😡
I will see to solve it. But definitely a real world case. Thanks for pointing to it.

thirtyinch/build.gradle
jreehuis: Having custom lint checks for a library is only really possible since AGP v3.0. Before you had to manually copy a lint.jar to specific folder and so on.
So not a lot libs ship their own lint checks with the AAR atm.

But in my view it is way more convenient if you have them out-of-the-box. If you do not want them, you could deactivate them in the Android Studio preferences anyway.

jreehuis:
> Something like this shouldn't be an issue:
> bildschirmfoto 2018-09-17 um 10 33 05
> 
> Ok, maybe this is not an issue because it will never happen in real world

In my view this is correct. StuffActivity could be also used separately and so we should highlight this error.

jreehuis:
> What does not work:
> 
> ```
> abstract class GalleryDetailsActivity : TiActivity<GalleryDetailsPresenter, GalleryDetailsView>()
> 
> class CameraGalleryDetailsActivity : GalleryDetailsActivity() {
> ```
> No Lint issue will be shown :(

OK, wasn't that hard than expected. This scenario should also be covered now.

jreehuis:
All done.

StefMa:
Issue created as discussed.
Merged ✅ 

Thank you @jreehuis 🤘

PR #160, opened 2018-09-12
Feature: getViewOrThrow() in onAttachView() Lint-Check

jreehuis:
## Description
This PR implements a new lint check to detect a direct or transitive call of `TiPresenter.getViewOrThrow()` in `TiPresenter.onAttachView()`.

In `TiPresenter.onAttachView()` the view of the presenter might still be null. So calling `TiPresenter.getViewOrThrow()` would throw an exception during runtime.
In production this usually crashes the App. This behaviour is correct to detect the issue later on.

With this lint check the issue can be easily detected already during development.

## Merge Information
Merge after #158 as this PR uses the Lint Check implementation of #158 as foundation.

### DIFF
--- a/thirtyinch-lint/src/main/kotlin/net/grandcentrix/thirtyinch/lint/TiIssue.kt
+++ b/thirtyinch-lint/src/main/kotlin/net/grandcentrix/thirtyinch/lint/TiIssue.kt
@@ -27,7 +27,7 @@ sealed class TiIssue(
 
     object GetViewOrThrowInOnAttach : TiIssue(
             id = "GetViewOrThrowInOnAttach",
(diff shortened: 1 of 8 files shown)


StefMa:
> In TiPresenter.onAttachView() the view of the presenter might still be null. 

How can this happen? 🤔 

> So calling TiPresenter.getViewOrThrow() would throw an exception during runtime.

Why should someone call `getViewOrThrow()` instead of using the given `View` from the method?

Do we need to update the **annotation** here - when its possible to have a `nullable` `View`? 
https://github.com/grandcentrix/ThirtyInch/blob/d761dbe800d935cd24a2e813ffd08ccdc4dddf61/thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiPresenter.java#L507

PR #161, opened 2018-09-17
add extension functions for boilerplate-less Disposable handling

lukaszkalnik:
Create Kotlin helper interface `DisposableHandler` to be implemented by `TiPresenter` to provide syntactic sugar for `Disposable` handling.

Implementing this interface by your Presenter enables you to call `.disposeWhenDestroyed()` and `.disposeWhenViewDetached()` as extension functions directly on your `Disposable` instances.

 Usage:
 ```
 class MyPresenter : TiPresenter<MyView>(), DisposableHandler {
 
    // configure with reference to your Presenter instance
     override val disposableHandler = RxTiPresenterDisposableHandler(this)
 
     override fun onCreate() {
         super.onCreate()
 
         // Presenter lifecycle dependent Disposable
         myObservable
             .subscribe()
             .disposeWhenDestroyed()
     }
 
     override fun onAttachView(view: MyView) {
         super.onAttachView(view)
 
         // View attached/detached dependent Disposable
         myViewObservable
             .subscribe()
             .disposeWhenViewDetached()
     }
 }
 ```


Syex:
What about for convenience adding a subclass of `TiPresenter` that implements the `DisposableHandler` for me? I'm sure we mostly have this class in our projects already and it would be the first thing I'd do in a project, because it adds that extra interface implementation and override as noise is every presenter otherwise.

The overhead when not needing the `disposableHandler` would be less important to me than having to implement and override this every time.

lukaszkalnik:
@Syex it's a good idea, although this interface was specifically meant as composition over inheritance, to not require a `BasePresenter` class to be extended by every Presenter.
But I will provide a convenience implementation of it.

jreehuis:
For our projects this really is a big win. But I do not like that we will include RxJava in each project which uses the Kotlin module. I guess there are projects whoch do not want to use RxJava but the Kotlin module.

lukaszkalnik:
@jreehuis then maybe we should put it in the thirtyinch-rx2 module? This is probably semantically more fitting, because indeed as you noticed it is related only to Rx usage in ThirtyInch.

lukaszkalnik:
I think most Android projects are moving to Kotlin sooner or later.

lukaszkalnik:
How can I retrigger the build? Espresso test timed out while waiting for window focus.

bildschirmfoto 2018-09-18 um 13 02 31

thirtyinch-rx2/build.gradle
```diff
@@ -23,6 +24,7 @@ android {
 
 dependencies {
     implementation project(':thirtyinch')
+    api "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
```
StefMa: 🙅‍♂️
Even if I agree with you, that Android projects starts more and more in Kotlin we should keep this to pure Java as long as we haven't discussed that internally first.

thirtyinch-rx2/build.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: 'com.android.library'
+apply plugin: 'kotlin-android'
```
StefMa: Please use `org.jetbrains.kotlin.android`.

thirtyinch-rx2/src/main/java/net/grandcentrix/thirtyinch/rx2/DisposableHandlingPresenter.kt
```diff
@@ -0,0 +1,16 @@
...
+import net.grandcentrix.thirtyinch.TiPresenter
+import net.grandcentrix.thirtyinch.TiView
+
+/**
+ * Convenience base [TiPresenter] class implementing [DisposableHandler] to be extended by your presenters.
+ *
+ * Makes possible to use [Disposable][io.reactivex.disposables.Disposable] handling extension functions without the
+ * need to implement the interface yourself.
+ */
+open class DisposableHandlingPresenter<V : TiView> : TiPresenter<V>(), DisposableHandler {
```
StefMa: ❌
I'm totally against this!
Please think from the "new user perspective":
* Why are there two Presenters?
* Do I have to use the DIsposableHandlingPresenter as soon as I use the rx2 dependency?

Something like this leads just to confusing instead of helping the user...

thirtyinch-rx2/build.gradle
lukaszkalnik: Good point. Then we have to probably put this PR on hold until it has been decided that we allow Kotlin code in ThirtyInch after all.

thirtyinch-rx2/src/main/java/net/grandcentrix/thirtyinch/rx2/DisposableHandlingPresenter.kt
lukaszkalnik: Yes, valid point. I was also not a fan of this, as it breaks the whole principle of composition over inheritance.

thirtyinch-rx2/build.gradle
lukaszkalnik: 👍

lukaszkalnik:
@passsy (and others), do we want this added or not?
If we don't want to add Kotlin support to ThirtyInch just for this small helper I think we should make a decision and close the PR. I'm fine with that.

lukaszkalnik:
I think the best solution would be to just publish it as a gist and link to it in the README.

lukaszkalnik:
Link to gist:
https://gist.github.com/lukaszkalnik/d9fb755021b385672fa4cfe29f6ccf81

PR #167, opened 2018-10-26
Cache gradle dir on CI

StefMa:
Don't know if it worth it.
Lets test it first...


StefMa:
It does not make sense :/
Travis/Android is slow. Deal with it

StefMa:
Gosh 🙄

PR #182, opened 2019-01-25
Add CoroutineScope for TiPresenter

Syex:
This PR adds a new module for Kotlin's coroutines which contains a custom `CoroutineScope` to be used when starting coroutines in a `TiPresenter`. The so started coroutines are then bound to the lifecycle of the presenter.

Mostly it will behave similar to the `RxTiPresenterDisposableHandler` introduced in #161, but instead of having to add `disposeWhenDestroyed` or `disposeWhenViewDetached` the `TiCoroutineScope` will cancel jobs automatically when the presenter reaches the specific lifecycle state.

If there's a better workaround to saving the last `presenterState` in a variable please let me know, maybe I just misunderstood the calls 🤷‍♂️


StefMa:
Kotlim 1.3.20 is already released

StefMa:
The Readme is wrong with the view detaching..
We have to add the boolean "true" als third param..
Beside if that. Does this really require to create each time a new view got
attached?

Syex:
> Beside if that. Does this really require to create each time a new view got
attached?

I was prepared for that question! Yes we have to. The `CoroutineScope`  requires its property `coroutineContext` to be final. We're initializing it with a `Job()`, which we cancel later. Once it is canceled we can't restart it and we can't reassign the `coroutineContext`. Do you see any other option?

Syex:
Changed the implementation. The flag, whether the scope should cancel all jobs when the view detaches, is no longer required to be constructor parameter.

Instead, there's now a method `launchUntilViewDetaches` on the scope which should be used for this case, rather than the default `launch` method.

README.md
```diff
@@ -271,6 +272,34 @@ interface HelloWorldView : TiView {
...
+class HelloWorldPresenter : TiPresenter<HelloWorldView> {
+
+  private val scope = TiCoroutineScope(this, Dispatchers.Default)
+
+  override fun onAttachView(view: HelloWorldView) {
+      scope.launchUntilViewDetaches { ... }
+  }
+}
+```
+However, be careful that `launchUntilViewDetaches` can only be called when there is a view attached!
```
StefMa: Can't we do it handle similar to the `deliverToView`? 🤔 
I mean currently it is like the `viewOrThrow`.
But in some cases we don't know if there is a view attached. Means, we call - for safety reasons - `deliverToView`.
Means we have to wrap the `launchUntilViewDetaches` into `deliverToView` to make sure its not crashing.
That is a bad behaviour and looks odd.

When we can handle it similar to `deliverToView` - execute when there is an view; otherwise wait until a view got attached - we don't have to wrap this into `deliverToView`

thirtyinch-kotlin-coroutines/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+    }
+    lintOptions {
+        abortOnError false
+    }
+}
+
+dependencies {
+    api project(":thirtyinch-kotlin")
+    api("org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version")
+    api("org.jetbrains.kotlinx:kotlinx-coroutines-android:1.0.1")
```
StefMa: 1.1.1 is already out

thirtyinch-kotlin-coroutines/build.gradle
```diff
@@ -0,0 +1,39 @@
...
+
+    testImplementation "junit:junit:$junitVersion"
+    testImplementation "com.nhaarman:mockito-kotlin:1.5.0"
```
StefMa: 2.1.0 is already out...

StefMa: But when you copy it from the other `kotlin` module keep it like it is 😁

thirtyinch-kotlin-coroutines/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
...
+#   public *;
+#}
+
+# Uncomment this to preserve the line number information for
+# debugging stack traces.
+#-keepattributes SourceFile,LineNumberTable
+
+# If you keep the line number information, uncomment this to
+# hide the original source file name.
+#-renamesourcefileattribute SourceFile
```
StefMa: Do we need this?
https://github.com/Kotlin/kotlinx.coroutines/blob/master/ui/kotlinx-coroutines-android/example-app/app/proguard-rules.pro

thirtyinch-kotlin-coroutines/src/main/java/net/grandcentrix/thirtyinch/kotlin/coroutines/TiCoroutineScope.kt
```diff
@@ -0,0 +1,79 @@
...
+import net.grandcentrix.thirtyinch.TiPresenter
+import net.grandcentrix.thirtyinch.TiPresenter.State.DESTROYED
+import net.grandcentrix.thirtyinch.TiPresenter.State.VIEW_ATTACHED
+import net.grandcentrix.thirtyinch.TiPresenter.State.VIEW_DETACHED
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * A [CoroutineScope] that is bound to the lifecycle of the given [TiPresenter].
+ *
+ * @param cancelWhenViewDetaches If all started coroutines in this scope should be cancelled when the view detaches.
```
StefMa: This is available anymore

thirtyinch-kotlin-coroutines/src/main/java/net/grandcentrix/thirtyinch/kotlin/coroutines/TiCoroutineScope.kt
```diff
@@ -0,0 +1,79 @@
...
+     * Parent [Job] for all coroutines that are started in the given presenter instance. When this `job` gets
+     * cancelled all of its children jobs will get cancelled too.
+     */
+    private val job = Job()
+    override val coroutineContext: CoroutineContext = context + job
+    /**
+     * Parent [Job] for all coroutines that are started while a `view` attached. When this `job` gets
+     * cancelled (when the view detaches) all of its children jobs will get cancelled too.
+     */
+    private var onViewDetachJob: Job? = null
```
StefMa: This Job is named (and documented) better as the "job" above.
I think it would be good to rename the `job` above to `presenterDestroyJob` and add the doc that this job will be destroyed when the presenter got destroyed...

thirtyinch-kotlin-coroutines/src/main/java/net/grandcentrix/thirtyinch/kotlin/coroutines/TiCoroutineScope.kt
```diff
@@ -0,0 +1,79 @@
...
+
+            when {
+                state == DESTROYED -> job.cancel()
+                state == VIEW_DETACHED && presenterState != null -> {
+                    onViewDetachCoroutineContext?.cancel()
+                    onViewDetachCoroutineContext = null
+                    onViewDetachJob = null
+                }
+                state == VIEW_ATTACHED -> {
+                    onViewDetachJob = Job(job)
```
StefMa: This doesn't have to be an child of the `job`, right?
Because it will be canceld in VIEW_DETACHED anyway.
And `VIEW_DETACHED` will be *always* called before `DESTROYED`.
Right?
But at the end it does not care :D

thirtyinch-kotlin-coroutines/src/main/java/net/grandcentrix/thirtyinch/kotlin/coroutines/TiCoroutineScope.kt
```diff
@@ -0,0 +1,79 @@
...
+                    onViewDetachCoroutineContext = context + onViewDetachJob!!
```
StefMa: We can use `apply{}` after the `Job` creation. Then we don't have to use the explanation marks...

thirtyinch-kotlin-coroutines/src/main/java/net/grandcentrix/thirtyinch/kotlin/coroutines/TiCoroutineScope.kt
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * Calling this method while no `view` is attached will throw a `IllegalStateException`, because no
+     * [CoroutineContext] is available at this time.
+     */
+    fun launchUntilViewDetaches(
+            block: suspend CoroutineScope.() -> Unit
+    ): Job {
+        val context = onViewDetachCoroutineContext ?: throw IllegalStateException(
+                "launchUntilViewDetaches can only be called when there is a view attached")
+        return launch(context = context, block = block)
```
StefMa: Is the named parameter required here? :D

README.md
Syex: No, what you describe is wrong. It has nothing to do with `viewOrThrow` or `deliverToView`. 

This should behave similar to [manageViewDisposable](https://github.com/grandcentrix/ThirtyInch/blob/master/thirtyinch-rx2/src/main/java/net/grandcentrix/thirtyinch/rx2/RxTiPresenterDisposableHandler.java#L95), which throws when there is now view. So, for the sake of consistency I made it the same.

thirtyinch-kotlin-coroutines/src/main/java/net/grandcentrix/thirtyinch/kotlin/coroutines/TiCoroutineScope.kt
Syex: Probably doesn't has to, but I think it's still more safe and reflects the hierarchy of the class.

thirtyinch-kotlin-coroutines/src/main/java/net/grandcentrix/thirtyinch/kotlin/coroutines/TiCoroutineScope.kt
Syex: Yes

PR #194, opened 2019-07-25
add testAttachView() extension function

lukaszkalnik:
This adds a convenience extension function which reduces tests boilerplate.
It is useful in typical tests where we always assume that the presenter has a view already attached and just want to test different presenter functions.

PR #195, opened 2019-07-26
Use dist trusty to build on travis

StefMa:
According to their (TravisCI) **current** documentation only "trusty" can build android projects:
Bildschirmfoto 2019-07-26 um 08 10 26

We don't used any specific ubuntu distribution. Therefore they picked (in the failing builds at #194) always an very old distribution:
Bildschirmfoto 2019-07-26 um 08 09 18

Hope for the best 🤞

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: android
-jdk:
-    - oraclejdk8
+dist: trusty
 
 android:
   components:


StefMa:
We get the same "report" as in #194.
Bildschirmfoto 2019-07-26 um 08 24 41
But this time travis simply ignore it and go through...

PR #166, opened 2018-10-26
Migrate to AndroidX

rehlma:

PR #197, opened 2019-08-09
Use AndroidX

StefMa:
Same as #166 but with a cleaner history 🙃.

Sry I was unable to do a rebase /w a force push at #166 and therefore I created a patch file, applied it and committed it by myself.
But I have you some credits in the commit message @rehlma ;)

I also removed all the "clutter" from #166 were we have done a lot more than "just updating to AndroidX"...
I will create new PRs with the stuff we did there 👍


thirtyinch/build.gradle
```diff
@@ -36,14 +36,15 @@ configurations {
 }
 
 dependencies {
-    compileOnly "com.android.support:appcompat-v7:$supportLibraryVersion"
+    compileOnly "androidx.appcompat:appcompat:$androidXAppCompatVersion"
+    compileOnly "androidx.annotation:annotation:$androidXAnnotationVersion"
 
-    testCompileOnly "com.android.support:support-annotations:$supportLibraryVersion"
+    testCompileOnly "androidx.annotation:annotation:$androidXAnnotationVersion"
```
rehlma: Didn't we say, it might makes more sense to use `implementation` here?

StefMa: Yep. But I just wanted to "migrate to androidx" and not "update all the things" 🙃

rehlma:
Ah okay I see. IMHO it was part of AndroidX. But it's fine : thumbsup:

Stefan M. <notifications@github.com> schrieb am Fr., 9. Aug. 2019, 14:41:

PR #199, opened 2019-08-12
Update RELEASE.md to use bundle*Aar task

### LINKED ISSUE #186: bundle[buildType] was replaced with AGP 3.2.0
We use `bundleRelease` to generate our AAR while creating a new release.
But this Task was renamed in the AGP 3.2.0.
We should use `bundleReleaseAar`.
-> https://github.com/grandcentrix/ThirtyInch/blob/master/RELEASE.md

The `bundle[XY]` tasks generates now only App Bundles. See https://developer.android.com/studio/build/building-cmdline#build_bundle & https://developer.android.com/guide/app-bundle

StefMa:
Fixes #186 

You can check this by doing the following:
First run `clean bundleRelease`.
Checkout the `thirtyinch/build/outputs/` dir. There is no **aar**  directory which contains the **aar**.

Now run the `bundleReleaseAar` task and check the same directory.
The aar directory is available and the aar file

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,11 +14,11 @@ Before we release a new version we have to make sure that we increase the `VERSI
 At this time we don't provide public `SNAPSHOT` releases.
 Anyway. If you want to use the latest and greatest features of Ti (or just want to build it by yourself because it is fun) you can "publish" Ti directly into your `mavenLocal()`. To do it you can run either:
 ```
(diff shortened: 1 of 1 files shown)


StefMa:
Rebased the current *master* (and forced push).
Otherwise the task wasn't available.

lukaszkalnik:
Now it works correctly.

PR #203, opened 2020-01-28
Increase AppCompat Version to 1.1.0 and JUnit to 4.13

jannisveerkamp:
There is a newer AppCompat version: 1.0.2 => 1.1.0
There is a newer JUnit version: 4.12 => 4.13


thirtyinch/src/main/java/net/grandcentrix/thirtyinch/TiActivity.java
```diff
@@ -121,7 +121,7 @@ protected void onStop() {
 
     @CallSuper
     @Override
-    protected void onSaveInstanceState(final Bundle outState) {
+    protected void onSaveInstanceState(@NonNull final Bundle outState) {
```
lukaszkalnik: There is no nullability annotation on this parameter in the appcompat library. Are you sure it's `@NonNull`?

jannisveerkamp: There is in version 1.1.0

PR #204, opened 2020-01-28
adjust version in readme and sample

lukaszkalnik:
Adjusts version to 1.0.0 in README and sample app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ ThirtyInch is available via [jcenter](http://blog.bintray.com/2015/02/09/android
 
 ```gradle
 dependencies {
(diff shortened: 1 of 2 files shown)


lukaszkalnik:
I don't know what's the legal practice on this one. All the classes still have 2016 as copyright year.
Should it always be updated with the current year of each release?

jannisveerkamp:
I think thats not required

lukaszkalnik:
Yes, it actually should stay as the date of first publication (and is not really required at all):
https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

PR #205, opened 2020-01-28
workaround for AndroidX Fragment 1.2.0 isInBackStack bug

lukaszkalnik:
There is a bug in AndroidX Fragment 1.2.0 `dump()` which caused our `isInBackStack()` hack to crash the app on configuration change in fragment or when switching fragments:
https://github.com/sockeqwe/mosby/issues/318#issuecomment-577660091

The issue is supposed to be fixed in AndroidX Fragment 1.2.1 so we will have to remove the workaround and restore the old solution then:
https://issuetracker.google.com/issues/148189412

Please test fragment configuration change and switching between fragments. App should not crash.


jannisveerkamp (approved):
Works 🙈

behi198:
Thanks!
after 24 hours working on this issue, at last I found this comment and updated and it works now!

